PR #4, opened 2026-04-09
ICP-2193 init WIP commit

armin-sams-dt:


thisthat (approved):

dt-ai-ingest/src/dt_ai_ingest/_sync.py
```diff
@@ -36,6 +36,10 @@ def _thread_target() -> None:
     thread.start()
     thread.join()
 
-    if isinstance(result, BaseException):
+    if isinstance(result, Exception):
         raise result
+    if isinstance(result, BaseException):
```
armin-sams-dt, line 41: Starts a thread, waits for it to finish, then re-raises any exception on the calling thread. Regular exceptions are re-raised as-is. Special exceptions like KeyboardInterrupt or SystemExit are re-raised without any extra Python error context attached, so they behave exactly as if no thread was involved at all.

dt-ai-ingest/src/dt_ai_ingest/_otel.py
```diff
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
```
armin-sams-dt, line 4: Token security: Access token is now stored as a SHA-256 hash only, preventing accidental leaks via logging or debug dumps.

Idempotency: The fast-path now checks provider object identity (_configured_provider is current), so if an external framework replaces the global provider between calls, the exporter is correctly re-attached instead of silently skipped.

dt-eval-lib/src/engine/evaluate.ts
```diff
@@ -110,14 +110,10 @@ async function callWithRetry(
   fn: () => Promise<LLMJudgeResponse>,
   maxRetries: number,
 ): Promise<LLMJudgeResponse> {
-  let lastError: unknown;
-
   for (let attempt = 0; attempt <= maxRetries; attempt++) {
```
armin-sams-dt, line 113: Retries fn up to maxRetries times on failure, with a sleep between attempts. Non-transient errors (EvalResponseError, EvalConfigError) are thrown immediately without retrying, since those won't resolve on their own. If all attempts fail, the last error is re-thrown.

PR #42, opened 2026-04-20
docs: update README for bedrock, azure-openai, and optional sampling

krismuhi:
## Summary

- Add `bedrock` and `azure-openai` to the Supported Providers table with their required env vars
- Add config file examples for both new providers
- Expand the environment variables section to cover AWS and Azure credentials
- Update `--sample` flag description to note it overrides config and that the default is random 5%
- Add inline comment in the config YAML example noting `sampling` is optional

## Test plan

- [ ] README renders correctly on GitHub
- [ ] All provider env var names match implementation in `dt-eval-cli/src/config/index.ts`
- [ ] Bedrock and Azure config examples match `bedrock-test.dt-eval.yaml` / `azure-openai-test.dt-eval.yaml`


yassine-dt (approved):
LGTM

PR #43, opened 2026-04-20
fix(dt-eval-lib): remove Node.js builtins via PromptStore DI

yassine-dt:
dt-eval-lib imported node:fs, node:os and node:path in custom.ts, preventing its use in Dynatrace frontend apps where these builtins are not available at bundle time.

- Add PromptStore interface and registerPromptStore() registry to dt-eval-lib
- Rewrite custom.ts to delegate to the registered store (no Node.js imports)
- listPrompts/getPrompt fall back gracefully (empty custom list) when no store registered
- createCustomPrompt/deleteCustomPrompt throw a clear error when no store registered
- Move filesystem logic into FsPromptStore class in dt-eval-cli
- Register FsPromptStore at CLI startup — existing CLI behavior unchanged

### DIFF
--- a/dt-eval-cli/src/index.ts
+++ b/dt-eval-cli/src/index.ts
@@ -1,5 +1,10 @@
 #!/usr/bin/env -S node --no-warnings
 
+import { registerPromptStore } from 'dt-eval-lib';
+import { FsPromptStore } from './prompts/fs-store.js';
+
+registerPromptStore(new FsPromptStore());
+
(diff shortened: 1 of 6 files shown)


krismuhi:
ok. this doesnt break my dt-eval evaluators add/delete  UX, so im fine with moving this out

thisthat (approved):

krismuhi (approved):

PR #46, opened 2026-04-21
feat(main): prompt to run eval immediately after setup wizard

krismuhi:
## Summary

- After the setup wizard saves the `.yaml` config file and prints the cost estimate, the user is now asked **"Would you like to run this right now?"** (default: yes)
- Selecting **yes** immediately runs `dt-eval run <config>` with full terminal output streamed live
- Selecting **no** prints the existing copy-paste hint as before

## Test plan

- [ ] Run `dt-eval configure` through the interactive wizard
- [ ] Verify the prompt appears after the cost estimate
- [ ] Confirm "yes" launches the run and output streams to the terminal
- [ ] Confirm "no" prints the copy-paste hint and exits cleanly
- [ ] Verify flag-only mode (`dt-eval configure --env-url ... --provider ...`) is unaffected

### DIFF
--- a/dt-eval-cli/src/cli/commands/configure.ts
+++ b/dt-eval-cli/src/cli/commands/configure.ts
@@ -1,4 +1,5 @@
 import { Command } from 'commander';
+import { spawnSync } from 'node:child_process';
 import { join, basename } from 'node:path';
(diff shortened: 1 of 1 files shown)


yassine-dt (approved):
LGTM. but fix the CI ahahah

thisthat (approved):

PR #51, opened 2026-04-29
fix(main): redact secretKey in configure --show output

krismuhi:
## Summary

- AWS Bedrock `secretKey` (secret access key) was printed in plaintext when running `dt-eval configure --show`
- Added `secretKey` redaction to `redactSecrets()` in `src/ui/format.ts`, consistent with the existing `apiKey` redaction

## Test plan

- [ ] Run `dt-eval configure --show` with a Bedrock config that has `secretKey` set — confirm it displays `***`
- [ ] Confirm `apiKey` still redacts as before


yassine-dt (approved):
LGTM

thisthat (approved):

PR #59, opened 2026-05-06
feat: platform tokens as primary auth, add dt-evals doctor command

krismuhi:
## Summary

- **Remove dtctl from the regular workflow** — `DT_API_TOKEN` (platform token) is now the only authentication method for `configure`, `run`, and all other commands. The `dtctlContext` config field and `DT_DTCTL_CONTEXT` env var are removed.
- **New `dt-evals doctor` command** — full environment health check and guided first-time setup.
- **New `dt-evals doctor create` subcommand** — focused flow to generate a platform token via dtctl OAuth and write it to `.env`.

## What `doctor` checks

| Section | Details |
|---------|---------|
| Dependencies | Node.js ≥18, dtctl installed and version |
| Authentication | Browser OAuth via `dtctl auth login` with spinner + continuous token polling (GitHub Copilot-style) |
| Permissions | `storage:spans:read`, `storage:events:read`, `storage:events:write`, `metrics:ingest` |
| Platform Token | Creates scoped API token via `/api/v2/apiTokens`, writes `DT_API_TOKEN` + `DT_ENV_URL` to `.env` |
| AI Provider | API key presence and connectivity (OpenAI, Anthropic, Azure, Gemini, Bedrock) |
| Config & Runs | Schema validation, last run status, 7-day failure rate |

The summary prints all errors and warnings with actionable remediation steps. If a token is missing it surfaces `dt-evals doctor create` as the next step.

## Changed files

- `src/dtctl/index.ts` *(new)* — isolated dtctl wrapper (context listing, OAuth browser flow with polling, permission probes, token creation)
- `src/cli/commands/doctor.ts` *(new)* — doctor command and `create` subcommand
- `src/dt/client.ts` — removed dtctl bearer auth; only `Api-Token` header remains
- `src/config/schema.ts` — removed `dtctlContext` field
- `src/config/index.ts` — removed `DT_DTCTL_CONTEXT` env var handling
- `src/cli/commands/configure.ts` — removed dtctl context prompt
- `src/cli/commands/run.ts` — removed dtctl auth fallback
- `src/ui/banner.ts` — added `printDoctorBanner()` (DT-EVALS-DOCTOR ASCII art)
- `README.md` — doctor usage, required Dynatrace token scopes table

## Test

```bash
cd dt-eval-cli
npm run dev -- doctor
npm run dev -- doctor create --context <your-dtctl-context>
npm run dev -- doctor --skip-auth   # config/provider/history only
```

PR #60, opened 2026-05-06
feat: configurable span field mapping and per-metric input routing

krismuhi:
## Summary

Two additive features for spans that don't follow OTel GenAI semconv, or
metrics that should score a subset of the conversation rather than the
joined transcript.

### `scope.spanFields` — map custom attributes to canonical fields

```yaml
scope:
  spanFields:
    input: llm.user_input
    output: [llm.response, custom.completion]
    systemInstruction: llm.system
    model: llm.model
```

User candidates are tried before built-in OTel + OpenLLMetry defaults
(first non-null wins), and the DQL `| fields` clause is built from the
deduplicated union of both — so the extra attributes are actually
fetched from Grail.

### `metrics.enabled[].inputs` — per-metric input routing

```yaml
metrics:
  enabled:
    - faithfulness                              # legacy string form
    - id: user-frustration
      inputs:
        input: userPrompt                       # last user-role prompt slot
```

Canonical fields: `input`, `output`, `systemInstruction`, `model`,
`userPrompt`. The new `userPrompt` is extracted by the parser from the
`gen_ai.prompt.N.role == "user"` slot (added to `GenAiSpan`). If the
requested field is empty on a span, the slot falls back to the default
so one misconfiguration doesn't silently drop spans.

## Schema bump 1 → 2

Both additions are strictly additive. v1 configs validate and run
unchanged against the v2 schema. `validateConfig` now also rejects
malformed `metrics.enabled` entries (must be string or `{ id, inputs? }`).

## Test plan

- [x] **138/138 unit tests pass** (+11 new)
  - 8 dql tests: `buildGenAiSpanQuery` includes user candidates,
    `parseSpanResults` prefers user attributes, falls back to defaults,
    extracts `userPrompt` from user-role slot
  - 3 runner tests: per-metric `inputs.input=userPrompt` routes the
    user turn into the eval, mixed string/object metric arrays,
    graceful fallback when canonical field is empty
- [x] `npx tsc --noEmit` adds 0 new errors (21 pre-existing on `main`,
      same count after change)
- [x] `npm run build` succeeds

## Notes for reviewers

- Helpers `metricId(entry)` and `metricInputs(entry)` keep call sites
  in `cli/commands/{run,status,configure}.ts` ergonomic when iterating
  `metrics.enabled`.
- `dt/dql.ts` was rewritten to walk a candidate list per slot rather
  than open-coded `??` chains — slightly more code, but the user
  candidate / built-in fallback story falls out naturally.
- Independent of the cross-tenant work in #59 / cross-tenant branch —
  this only touches parsing and runner input shaping.

### DIFF
--- a/dt-eval-cli/README.md
+++ b/dt-eval-cli/README.md
@@ -383,6 +383,58 @@ Available canonical fields: `input`, `output`, `systemInstruction`, `model`,
 given span, the slot falls back to `span.input` / `span.output` /
 `span.systemInstruction` so a single misconfiguration doesn't drop spans.
 
+### Full example
+
+A complete `.dt-eval.yaml` combining everything — schema bump, custom
+span field mapping, per-metric input routing, sampling, alerts:
+
+```yaml
+schemaVersion: 2
+name: travel-assistant-prod
+
+dynatrace:
+  environmentUrl: https://your-env.live.dynatrace.com
+  # apiToken loaded from DT_API_TOKEN env var
+
+judge:
+  provider: azure-openai
+  model: gpt-4.1-mini
+  # apiKey, baseUrl, apiVersion loaded from AZURE_OPENAI_* env vars
+
+scope:
+  service: travel-assistant
+  since: 1h
+  sampling:
+    strategy: latest
+    count: 50
+  # Map custom span attributes to canonical fields. Defaults handle OTel
+  # GenAI semconv + OpenLLMetry; override only what you need.
+  spanFields:
+    output: [gen_ai.output.message, gen_ai.output.messages]
+    # input, systemInstruction, model use built-in defaults
+
+metrics:
+  enabled:
+    # Plain string form — uses span.input / span.output / span.systemInstruction
+    - faithfulness
+    - relevance
+    - hallucination
+    - answer-completeness
+    # Object form — overrides which canonical span field feeds the
+    # evaluator's `input` slot. user-frustration scores the user's turn
+    # alone instead of the full joined transcript.
+    - id: user-frustration
+      inputs:
+        input: userPrompt
+
+alerts:
+  thresholds:
+    faithfulness: 0.7
+    relevance: 0.7
+    answer-completeness: 0.8
+    user-frustration: 1
+```
+
 Useful environment variables:
 
 ```bash
--- a/dt-eval-cli/README.md
+++ b/dt-eval-cli/README.md
@@ -275,7 +275,7 @@ Config is resolved in this order:
 ### Example config
 
 ```yaml
-schemaVersion: 1
+schemaVersion: 2
 name: travel-assistant-prod
 
 dynatrace:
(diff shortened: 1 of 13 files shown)


krismuhi:
there will be a followup PR, once doctor is merged, to automatically detect this from a fetched span, intelligent mapping based on the returned gen_ai span data

dt-eval-cli/src/dt/dql.ts
```diff
@@ -25,32 +54,66 @@ export function buildGenAiSpanQuery(opts: DqlQueryOptions): string {
   lines.push(`| filter start_time > now() - ${since}`);
 
   if (app) {
-    // Match by service.name or the entity ID alternative
-    lines.push(`| filter service.name == "${app}" or dt.smartscape.service == "${app}"`);
```
thisthat, line 29: we could use `toSmartscapeId(value)`: [docs](https://docs.dynatrace.com/docs/platform/grail/dynatrace-query-language/functions/conversion-and-casting-functions#toSmartscapeId)

dt-eval-cli/src/dt/dql.ts
```diff
@@ -25,32 +54,66 @@ export function buildGenAiSpanQuery(opts: DqlQueryOptions): string {
...
+    // Match by service.name only — `dt.smartscape.service` holds an entity
+    // ID (e.g. `SERVICE-1A2B…`), not a service name, so comparing it to a
+    // string raises a SMARTSCAPEID_TO_STRING_COMPARISON warning from Grail
+    // and matches nothing.
+    lines.push(`| filter service.name == "${app}"`);
```
thisthat: dunno if we should also check for `dt.service.name` [docs](https://docs.dynatrace.com/docs/semantic-dictionary/fields#dynatrace-service-fields)

dt-eval-cli/src/dt/dql.ts
```diff
@@ -25,32 +54,66 @@ export function buildGenAiSpanQuery(opts: DqlQueryOptions): string {
...
+  // Build a deduplicated field list. `gen_ai.system` and `gen_ai.provider.name`
+  // are needed for the system filter; the rest are built-in + user candidates.
+  const fieldSet = new Set<string>([
+    'trace.id',
+    'span.id',
+    'start_time',
+    'status.code',
+    'gen_ai.system',
+    'gen_ai.provider.name',
+    ...fields.input,
```
thisthat, line 91: we lost `gen_ai.request.model` and `gen_ai.completion.0.content`

dt-eval-cli/README.md
```diff
@@ -319,6 +319,52 @@ Sampling strategies:
...
+    - faithfulness                                  # legacy string form
+    - id: user-frustration
+      inputs:
+        input: userPrompt                           # latest user-role prompt slot
+    - id: hallucination
+      inputs:
+        context: systemInstruction                  # use system prompt as context
+```
+
+Available canonical fields: `input`, `output`, `systemInstruction`, `model`,
```
thisthat, line 380: maybe we should provide a full example

dt-eval-cli/src/dt/dql.ts
krismuhi: Both still in the query — they live in the deduplicated default field list:

```ts
const DEFAULT_OUTPUT_FIELDS = ['gen_ai.output.message', 'gen_ai.completion.0.content'];
const DEFAULT_MODEL_FIELDS  = ['gen_ai.request.model'];
```

`buildGenAiSpanQuery` builds a `Set<string>` from the union of user candidates and these defaults, then renders `| fields …` from that Set. Tests assert both fields are emitted (`tests/dt/dql.test.ts` — "includes required fields in the fields clause"). The hunk you're reading on line 83 is just the rendering step; the values come from those constants a few lines up.

dt-eval-cli/src/dt/dql.ts
krismuhi: done

dt-eval-cli/src/dt/dql.ts
krismuhi: done

dt-eval-cli/README.md
krismuhi: added

PR #61, opened 2026-05-06
feat: cross-tenant origin/destination support

krismuhi:
> Built on top of #59 — review after that lands. Once #59 merges, this PR's diff vs `main` will be the cross-tenant layer only.

## Summary

Splits Dynatrace connectivity into two endpoints — `origin` (DQL reads,
typically the production tenant where GenAI spans live) and `destination`
(bizevent + metric writes, typically a separate eval-results tenant).
Single-tenant configs keep working unchanged.

```yaml
dynatrace:
  origin:
    environmentUrl: https://prod.live.dynatrace.com
    apiToken: dt0s16.xxxxx          # storage:spans:read, storage:buckets:read
  destination:
    environmentUrl: https://eval-results.dev.apps.dynatracelabs.com
    apiToken: dt0s16.yyyyy          # storage:bizevents:write, storage:metric:write
```

Top-level `environmentUrl` / `apiToken` (when present) fall back into either side, so existing configs are equivalent to ``origin == destination == top-level``.

## Changes

- **`config/schema.ts`** — `DynatraceConfig extends Partial<DynatraceEndpoint>`, adds optional `origin` / `destination`. Exports `resolveEndpoints()` which produces two concrete endpoints with top-level fallback.
- **`config/index.ts`** — env vars `DT_ORIGIN_ENV_URL` / `DT_ORIGIN_API_TOKEN` / `DT_DESTINATION_ENV_URL` / `DT_DESTINATION_API_TOKEN` flow to the matching side at highest precedence. `validateConfig` reports per-side URL issues.
- **`runner/index.ts`** — origin client for `executeDql`; destination client for `ingestBizevents` / `ingestMetrics`. Same client for both in single-tenant configs.
- **`cli/commands/{run,validate,status,schedule}.ts`** — surfaces both sides consistently. `validate` now runs a real ``fetch spans | limit 1`` against origin via the new `client.probeBucketRead('spans')`, catching `MISSING_BUCKET_PERMISSIONS` warnings that Grail returns as SUCCEEDED-with-empty-records.
- **`dt/client.ts`** — adds `probeBucketRead(table)` for the validate-side scope check; switches bizevent / metric ingest URLs to the ``/platform/classic/environment-api/v2/...`` surface (broadly available across tenant generations and explicitly tested working on the cross-tenant target).

## Test plan

- [x] **136/136 unit tests pass** (+6 new)
  - `resolveEndpoints` top-level fallback, per-side override precedence
  - env-var precedence (`DT_ORIGIN_API_TOKEN` / `DT_DESTINATION_API_TOKEN`)
  - `validateConfig` accepts an origin/destination-only config (no top-level url)
  - `probeBucketRead` ok-path and `MISSING_BUCKET_PERMISSIONS` failure path
  - Existing bizevent URL test updated for the new ingest endpoint
- [x] `npx tsc --noEmit` adds **0 new errors** (24 errors total, all pre-existing — same set as main, with three line numbers shifted due to new code in validate.ts and runner/index.ts)
- [x] `npm run build` succeeds
- [x] **Manual end-to-end run** against a real cross-tenant pair (production read tenant → dev write tenant): 5 GenAI spans fetched, 5 evaluations dispatched to the judge, 5 bizevents land in the destination tenant queryable via DQL.

## Independent of #60

This PR and #60 (configurable span field mapping) touch disjoint files except for the runner's call site. Either can land first; the merge order doesn't matter.

### DIFF
--- a/dt-eval-cli/src/dt/client.ts
+++ b/dt-eval-cli/src/dt/client.ts
@@ -69,12 +69,33 @@ function logDqlNotifications(metadata: DqlResultMetadata | undefined): void {
 const POLL_INTERVAL_MS = 1000;
 const POLL_TIMEOUT_MS = 60_000;
 
+/**
+ * Translate `*.apps.dynatrace.com` → `*.live.dynatrace.com`.
+ *
+ * Modern Dynatrace platform splits its surface across two subdomains:
+ *   - `.apps.` hosts the platform-storage / DQL APIs (`/platform/storage/...`).
+ *   - `.live.` hosts the classic environment APIs (`/api/v2/...`).
+ *
+ * Users supply a single `environmentUrl` for the tenant — by convention the
+ * `.apps.` form (because DQL is the entry point). For ingest endpoints
+ * (`/api/v2/bizevents/ingest`, `/api/v2/metrics/ingest`) we need to redirect
+ * to `.live.`, otherwise the apps gateway returns 400 *Invalid app context*.
+ *
+ * Idempotent on URLs that already point at `.live.` or any other host
+ * (cluster-managed, on-prem, dev sandboxes), so it's safe to call always.
+ */
+function liveSubdomain(url: string): string {
+  return url.replace(/^(https?:\/\/[^/]+?)\.apps\.dynatrace\.com/, '$1.live.dynatrace.com');
+}
+
 export class DynatraceClient {
   private readonly environmentUrl: string;
+  private readonly liveBaseUrl: string;
   private readonly apiToken: string;
 
   constructor(config: DynatraceClientConfig) {
     this.environmentUrl = config.environmentUrl.replace(/\/$/, '');
+    this.liveBaseUrl = liveSubdomain(this.environmentUrl);
     this.apiToken = config.apiToken ?? '';
   }
 
@@ -152,7 +173,8 @@ export class DynatraceClient {
   }
 
   async ingestBizevents(events: object[]): Promise<void> {
-    const url = `${this.environmentUrl}/platform/classic/environment-api/v2/bizevents/ingest`;
+    // Classic env-api on `.live.` subdomain. See `liveSubdomain()` for why.
+    const url = `${this.liveBaseUrl}/api/v2/bizevents/ingest`;
     const response = await fetch(url, {
       method: 'POST',
       headers: this.authHeaders(),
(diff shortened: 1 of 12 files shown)


dt-eval-cli/tests/dt/client.test.ts
```diff
@@ -153,7 +153,7 @@ describe('DynatraceClient', () => {
 
     expect(mockFetch).toHaveBeenCalledOnce();
     const [url, init] = mockFetch.mock.calls[0] as [string, RequestInit];
-    expect(url).toBe('https://test.live.dynatrace.com/platform/ingest/v1/bizevents');
+    expect(url).toBe('https://test.live.dynatrace.com/platform/classic/environment-api/v2/bizevents/ingest');
```
thisthat: 🤔 
it should be 
```suggestion
    expect(url).toBe('https://test.live.dynatrace.com/api/v2/bizevents/ingest');
```

dt-eval-cli/src/dt/client.ts
```diff
@@ -119,8 +119,40 @@ export class DynatraceClient {
...
+    const notifications = data.result?.metadata?.grail?.notifications ?? [];
+    const missing = notifications.find(n => n.notificationType === 'MISSING_BUCKET_PERMISSIONS');
+    if (missing) {
+      return { ok: false, reason: missing.message ?? 'MISSING_BUCKET_PERMISSIONS' };
+    }
+    return { ok: true };
+  }
+
   async ingestBizevents(events: object[]): Promise<void> {
-    const url = `${this.environmentUrl}/platform/ingest/v1/bizevents`;
```
thisthat, line 123: afaik the original URL is correct but it should be `{environmentUrl}/api/v2/bizevents/ingest`

krismuhi: You're right — switched to `/api/v2/bizevents/ingest` (and the matching `/api/v2/metrics/ingest`) in `ef863a0`.

Caveat: that path only resolves on the `.live.` subdomain, while DQL lives on `.apps.`. Users supply a single `environmentUrl` (by convention the `.apps.` form, since it's the platform entry point), so the client now rewrites `*.apps.dynatrace.com` → `*.live.dynatrace.com` for the two ingest calls (DQL stays on `.apps.`). Idempotent on managed / on-prem / dev-sandbox URLs.

Probed both endpoint forms with a platform-token Bearer on a clean prod tenant after the IAM policy was bound:

| URL | result |
|---|---:|
| `apps + /platform/classic/...` | 202 ✅ |
| `live + /api/v2/...` | **202 ✅** |
| `apps + /api/v2/...` | 400 *Invalid app context* |
| `live + /platform/classic/...` | 403 *administrative rules* |

Both paired forms work; `/api/v2/...` is the documented public surface, so going with that. Doctor probes follow the same translation.

krismuhi: fixed it. found the issue

dt-eval-cli/tests/dt/client.test.ts
krismuhi: fixed

PR #88, opened 2026-05-13
feat(dt-eval-cli): make the run spinner reflect the actual phase

krismuhi:
## Summary

The spinner during `dt-evals run` was hardcoded to `Fetching GenAI spans from Dynatrace...` for the entire run. The actual DQL fetch is a single call (usually <500ms); the rest of the time is the LLM judge calls. So the spinner was lying about what was happening — users saw `Fetching...` for 30+ seconds while it was really evaluating.

This PR adds a `RunProgressEvent` stream to `runEvals` and drives the spinner from it. Now you see:

```
✓ Fetched 9 spans in 234ms
⠙ Evaluating 12/27 • relevance trace=685603c1…
✓ Evaluated 27 tasks in 12.3s
✓ Wrote 27 results in 180ms
```

## What changed

- `src/runner/index.ts`: new `RunProgressEvent` discriminated union and optional `onProgress` callback on `RunOptions`. Events fire at each phase boundary (`fetching`, `fetched`, `preparing`, `evaluating-start`, `eval-completed`, `evaluating-done`, `writing`, `written`) with span counts and per-phase timing. The eval-loop handler now also surfaces per-task completion (including the metric and trace ID) so the spinner can render live N/M progress.
- `src/cli/commands/run.ts`: new `buildSpinnerProgress` helper maps events to `spinner.update` / `spinner.succeed`. The previous `spinner.succeed("Evaluation run complete...")` at the end is replaced with a `stop()` — each phase already emitted its own success line.
- `tests/runner/index.test.ts`: new test asserts the full phase sequence fires in order and per-task `eval-completed` count matches the task fan-out.

## Backwards compatibility

`onProgress` is optional. When omitted, the runner falls back to the existing `logger.step` / `logger.info` phase markers, so non-CLI callers (tests, library use, scheduled runs) see no behavior change.

## Test plan

- [x] `npm test` — 159/159 passing (158 prior + 1 new)
- [x] `npx tsc --noEmit` — clean
- [ ] manual: run with `--debug` and verify each phase prints `✓ Fetched ... in Xms`, live `Evaluating N/M …` updates, `✓ Wrote ...`
- [ ] manual: `--ci` mode still produces the JSON-only output (no spinner, no progress lines)
- [ ] manual: `--dry-run` returns cleanly without an in-flight spinner line dangling


krismuhi:
reviewed and fixed with Codex

Tested against prod:
image
image

PR #100, opened 2026-06-01
feat(dt-eval-cli): server-side eval alerts as Dynatrace Workflows

krismuhi:
## Summary

Adds `dt-evals alerts` — a new command tree that deploys per-notification Dynatrace Workflows. Each workflow polls eval bizevents on a schedule and notifies Slack / email / a generic webhook when a threshold is breached. Sits next to the existing local `alerts.thresholds` / `alerts.webhooks` blocks; this is the continuous server-side path.

## Why

Today `alerts.webhooks` fires only inside a local `dt-evals run`, so it misses every other run (CI, scheduled jobs, drift baselines). This change moves alerting into Dynatrace itself so it works across all eval activity, with the same telemetry the dashboards already consume.

## YAML

```yaml
alerts:
  notifications:
    - name: toxicity-spike
      metric: toxicity
      condition: avg > 2           # avg | max | min | count | fail_rate
      window: 15m                  # also the polling interval
      channel:
        type: slack
        connection: my-slack-conn  # Dynatrace Slack OAuth connection
        channel: "#ai-alerts"
```

## Commands

| Command | Purpose |
|---|---|
| `alerts apply  [config]` | idempotent create/update (per-config workflow id map at `~/.dt-eval/alerts-<hash>.json`) |
| `alerts render [config]` | print rendered workflow YAML, no API calls |
| `alerts list   [config]` | list deployed alerts (`--all` for tenant-wide) |
| `alerts diff   [config]` | local vs remote, normalised through JSON to strip server-injected fields |
| `alerts delete [config] --name X \| --all` | per-alert or full teardown |

## Channels

- `type: slack` → `dynatrace.slack:slack-send-message`. Requires a Dynatrace Slack connection (Settings → Connections → Slack OAuth). Renders the proper Slack action UI in the Workflows app.
- `type: email` → `dynatrace.email:send-email`. Uses the tenant's default outbound mail (Settings → Notifications); no per-workflow connection.
- `type: webhook` → `dynatrace.automations:run-javascript` that POSTs JSON via `fetch()`. Core AutomationEngine action; no extra app install required. Useful for Slack incoming webhooks, PagerDuty proxies, custom routers.

## Condition DSL

`<aggregation> <operator> <value>[%]` parsed at validate time and translated to DQL.

- aggregations: `avg`, `max`, `min`, `count` (failing rows), `fail_rate`
- operators: `>`, `>=`, `<`, `<=`, `==`, `!=`
- `fail_rate` is a percentage — write `fail_rate > 10%` or `fail_rate > 0.1`. Bare values outside `[0, 1]` are rejected to avoid the 10-vs-10% trap. Rendered alert messages always show fail_rate as a percentage regardless of which form was used.

## Permissions

Required dtctl scopes (`automation:workflows:read`, `automation:workflows:write`, `storage:events:read`, plus optional `automation:workflows:run`) are surfaced in the `alerts --help` description and printed on every 401/403 from dtctl. After every `apply` / `list`, a deep link to the Workflows app filtered to dt-evals is printed; the URL follows `--context` so it points at whichever tenant was deployed to.

## Tested end-to-end against a live tenant

Verified every command against a live tenant:

- `render` (stdout + `--out <dir>`), `apply` (fresh + idempotent re-run + `--dry-run` + `--name`), `list` (scoped + `--all`), `diff` (in-sync + change detection), `delete` (`--name` + `--all` + missing-name warning + 404-stale handling)
- Slack notify task loads the proper form UI in the Workflows app (Connection / Channel / Message / Reactions / Reply in thread)
- Email notify task uses the working `to / cc / bcc / subject / content` shape
- Generic webhook notify task uses the always-available `run-javascript` core action — no `AppNotFound` failures

## Files

- `src/alerts/{condition,render,idmap,dtctl}.ts` — condition DSL, workflow renderer, idempotency state, dtctl wrappers with classified 401/403 errors
- `src/cli/commands/alerts.ts` — `apply / render / list / diff / delete`
- `src/config/{schema,index}.ts` — `alerts.notifications[]` type + validation (including condition-DSL parse at validate time)
- `tests/alerts/{condition,render}.test.ts` — 25 unit tests covering the DSL, percentage trap, all channel shapes, link rendering, idempotent id behaviour
- `alerts-examples.dt-eval.yaml` — five realistic patterns (PII / prompt-injection page-on-any, relevance regression email, toxicity-spike Slack, hallucination-trend generic webhook)
- `docs/alerts-design.md` — design doc captured from the planning round
- README: new sections for the YAML block, condition DSL, channel setup, permissions, tenant-URL print-out

## Test plan

- [ ] `dt-evals alerts render alerts-examples.dt-eval.yaml` prints five workflow YAMLs
- [ ] `dt-evals alerts apply alerts-examples.dt-eval.yaml --dry-run` shows the action table without API calls
- [ ] `dt-evals alerts apply alerts-examples.dt-eval.yaml --context <tenant>` creates five workflows, prints the tenant URL, and the connection callout
- [ ] Re-running `apply` updates in place (no duplicate workflows in the UI)
- [ ] Editing a `window:` in the YAML and running `alerts diff` shows only the three actually-changed lines (DQL `from:`, cron, human description)
- [ ] `alerts delete --name X` removes one and clears the id-map entry; `--all` tears down the rest
- [ ] On a tenant where the Slack connection name doesn't exist, opening the `notify` task in the UI still shows the proper Slack action form (just with no connection selected)

### DIFF
--- a/dt-eval-cli/alerts-examples.dt-eval.yaml
+++ b/dt-eval-cli/alerts-examples.dt-eval.yaml
@@ -11,7 +11,7 @@ schemaVersion: 2
 name: pydantic-music-agent
 
 dynatrace:
-  environmentUrl: https://mho70695.apps.dynatrace.com
+  environmentUrl: ${DT_ENV_URL}
 
 judge:
   provider: openai
--- a/dt-eval-cli/alerts-examples.dt-eval.yaml
+++ b/dt-eval-cli/alerts-examples.dt-eval.yaml
@@ -1,0 +1,96 @@
+# Example alerts config — five realistic patterns you can apply with:
+#
+#   dt-evals alerts apply alerts-examples.dt-eval.yaml
+#
+# Each notification demonstrates a different aggregation/window/channel mix.
+# The `connection:` names below are placeholders — to actually deliver, either
+# create connections with those names in Dynatrace, or swap them for real
+# Slack incoming-webhook URLs under `webhookUrl:`.
+
+schemaVersion: 2
+name: pydantic-music-agent
+
+dynatrace:
+  environmentUrl: https://mho70695.apps.dynatrace.com
+
+judge:
+  provider: openai
+  apiKey: ${OPENAI_API_KEY}
+
+scope:
+  service: pydantic-ai-music-agent      # ~3.4k eval bizevents in last 30d
+  since: 1h
+  sampling: { strategy: random, percent: 5 }
+
+metrics:
+  enabled: [toxicity, relevance, hallucination, pii-leakage, prompt-injection]
+
+alerts:
+  notifications:
+
+    # ──────────────────────────────────────────────────────────────────────
+    # 1) Critical safety: page on ANY PII leak in the last 5 minutes.
+    #    Tight window, zero-tolerance threshold, Slack so it's noisy.
+    # ──────────────────────────────────────────────────────────────────────
+    - name: pii-leak-any
+      metric: pii-leakage
+      condition: count > 0           # any failing pii-leakage eval at all
+      window: 5m
+      channel:
+        type: slack
+        connection: ai-oncall-slack
+        channel: "#ai-oncall"
+
+    # ──────────────────────────────────────────────────────────────────────
+    # 2) Critical safety: same shape for prompt injection.
+    # ──────────────────────────────────────────────────────────────────────
+    - name: prompt-injection-any
+      metric: prompt-injection
+      condition: count > 0
+      window: 5m
+      channel:
+        type: slack
+        connection: ai-oncall-slack
+        channel: "#ai-oncall"
+
+    # ──────────────────────────────────────────────────────────────────────
+    # 3) Quality regression: more than 20% of relevance evals failed in 1h.
+    #    Slower window since one bad batch shouldn't page anyone; email so it
+    #    lands in the AI team inbox for triage instead of a paging channel.
+    # ──────────────────────────────────────────────────────────────────────
+    - name: relevance-regression
+      metric: relevance
+      condition: fail_rate > 20%
+      window: 1h
+      channel:
+        type: email
+        # Uses the tenant's default outbound mail config (Dynatrace → Settings →
+        # Notifications). No per-workflow connection needed.
+        to: [ai-quality@example.com]
+        subject: "[dt-evals] relevance regression on pydantic-ai-music-agent"
+
+    # ──────────────────────────────────────────────────────────────────────
+    # 4) Toxicity spike on the 1–5 scale: average above 2 for 15 minutes.
+    #    Routed to a different Slack channel than the on-call ones above.
+    # ──────────────────────────────────────────────────────────────────────
+    - name: toxicity-spike
+      metric: toxicity
+      condition: avg > 2
+      window: 15m
+      channel:
+        type: slack
+        connection: ai-oncall-slack
+        channel: "#ai-quality"
+
+    # ──────────────────────────────────────────────────────────────────────
+    # 5) Hallucination trend: more than 10% failing over a 1-hour window.
+    #    Generic webhook so an external router (PagerDuty proxy, custom
+    #    on-call tool, etc.) can consume the structured JSON payload.
+    # ──────────────────────────────────────────────────────────────────────
+    - name: hallucination-trend
+      metric: hallucination
+      condition: fail_rate > 10%
+      window: 1h
+      channel:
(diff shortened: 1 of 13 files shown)


dt-eval-cli/alerts-examples.dt-eval.yaml
```diff
@@ -0,0 +1,96 @@
...
+# Each notification demonstrates a different aggregation/window/channel mix.
+# The `connection:` names below are placeholders — to actually deliver, either
+# create connections with those names in Dynatrace, or swap them for real
+# Slack incoming-webhook URLs under `webhookUrl:`.
+
+schemaVersion: 2
+name: pydantic-music-agent
+
+dynatrace:
+  environmentUrl: https://mho70695.apps.dynatrace.com
```
thisthat: ${DT_ENV} or what's the var name used to configure the eval

krismuhi: fixed.

dt-eval-cli/src/alerts/condition.ts
```diff
@@ -0,0 +1,114 @@
...
+  /** True if the user wrote a percentage (e.g. "10%"); only meaningful for fail_rate. */
+  isPercent: boolean;
+  /** The original DSL string, for error messages and rendered descriptions. */
+  raw: string;
+}
+
+const VALID_AGGS: Aggregation[] = ['avg', 'max', 'min', 'count', 'fail_rate'];
+const VALID_OPS: Operator[] = ['>=', '<=', '==', '!=', '>', '<'];
+
+const CONDITION_RE = /^\s*(avg|max|min|count|fail_rate)\s*(>=|<=|==|!=|>|<)\s*(-?\d+(?:\.\d+)?)\s*(%?)\s*$/i;
```
egehandulger, line 28: Clause says: The -? allows negative values. count > -1 would always fire. No downstream guard catches this.

dt-eval-cli/src/alerts/render.ts
```diff
@@ -0,0 +1,335 @@
...
+  if (n <= 0) throw new Error(`Window must be > 0 (got "${window}")`);
+
+  const humanUnit: Record<typeof unit, string> = { s: 'second', m: 'minute', h: 'hour', d: 'day' };
+  let cron: string;
+  switch (unit) {
+    case 's':
+    case 'm': {
+      const minutes = unit === 's' ? Math.max(1, Math.round(n / 60)) : n;
+      cron = minutes >= 60 ? `0 */${Math.floor(minutes / 60)} * * *` : `*/${minutes} * * * *`;
+      break;
```
egehandulger, line 76: I understand we don't want this to run more frequent than once a minute so we set the cron accordingly.
If `20s` is passed, we set it to `1m` in the cron however, the DQL still uses `20s` as a window so `40s` of data is not analyzed. Is this desired?

dt-eval-cli/src/cli/commands/alerts.ts
```diff
@@ -0,0 +1,460 @@
...
+  const localLines = local.split('\n');
+  const max = Math.max(remoteLines.length, localLines.length);
+  for (let i = 0; i < max; i++) {
+    const r = remoteLines[i];
+    const l = localLines[i];
+    if (r === l) continue;
+    if (r !== undefined) console.log(`- ${r}`);
+    if (l !== undefined) console.log(`+ ${l}`);
+  }
+}
```
egehandulger, line 444: This is a very simple diffing algorithm. If a new line is added somewhere, all the remaining lines would seem changed. Maybe we utilize a library that does this more complex.

dt-eval-cli/src/cli/commands/alerts.ts
```diff
@@ -0,0 +1,460 @@
...
+        const tmp = join(tmpdir(), `dt-evals-alert-${n.name}-${Date.now()}.yaml`);
+        writeFileSync(tmp, rendered.yaml, 'utf-8');
+        rows.push([n.name, existingId ? 'would update' : 'would create', existingId ?? '-', tmp]);
+        continue;
+      }
+
+      try {
+        logger.step(`apply ${n.name}${existingId ? ` (update ${existingId})` : ' (create)'}`);
+        const id = await applyWorkflow(rendered.yaml, { context: opts.context });
+        updated = setMapping(updated, n.name, id);
```
egehandulger, line 236: This takes into account all the IDs that are new and sets them but how about the ones removed? `removeMapping` should also be called?

PR #104, opened 2026-06-03
feat(dt-eval-cli): split metric into metricId/metricName in bizevent payload

egehandulger:
bizevents would now have both metric id and the human readable metric name


thisthat (approved):
LGTM :shipit:

krismuhi:
just noting down: breaks the ready-made dashboards that i need to update

krismuhi (approved):

PR #112, opened 2026-06-12
fix(dt-eval-cli): correct token scopes in README and parse gen_ai.output.messages

krismuhi:
## What

Two independent fixes found while running `dt-evals` end-to-end against a real customer instance (`mho70695.apps.dynatrace.com`).

### 1. README scope names are wrong and incomplete

The cross-tenant config example had two typos and two missing scopes compared to what the code actually uses (`dtctl/index.ts`, `doctor.ts`):

| README said | Code actually uses |
|---|---|
| `storage:bizevents:write` | `storage:events:write` |
| `storage:metric:write` | `storage:metrics:write` |
| *(missing)* | `storage:buckets:read` |
| *(missing)* | `storage:logs:read` |

`storage:buckets:read` is particularly dangerous to leave undocumented — without it Grail returns `SUCCEEDED` with empty records and no error, so spans silently disappear and `dt-evals run` reports 0 results with no indication of what went wrong.

Adds a dedicated **Required token scopes** table so all scopes (including the alerts/workflows ones) are in one place rather than scattered across examples.

### 2. `gen_ai.output.messages` (plural) not in default output candidates

When testing against the instance, `dt-evals run` fetched 0 spans despite 103 GenAI spans existing in Dynatrace. Running with `--debug --verbose` showed the DQL query was correct and returning records. The spans were being silently dropped in `parseSpanResults` at the `if (!input || !output) continue` guard (`dql.ts:243`).

Root cause: the instance uses the OTel GenAI evolved convention where the full conversation — including the assistant response — is stored as a JSON array under `gen_ai.output.messages` (plural). The default output candidate list only contained `gen_ai.output.message` (singular) and `gen_ai.completion.0.content`. The parser already handles the JSON array role extraction via `extractRolesFromJsonMessages`, the field just wasn't being fetched.

Fix: add `gen_ai.output.messages` to `DEFAULT_OUTPUT_FIELDS` so it is included in the DQL `fields` clause and checked before falling back to the singular form.

Test added to `dql.test.ts` covering a span with `gen_ai.output.message: null` and `gen_ai.output.messages` containing a role-tagged assistant message.

## Test plan
- [x] `npm test` — 161/161
- [x] Live run against `mho70695.apps.dynatrace.com` without `spanFields` workaround: 103 spans fetched, 10 eval results written

### DIFF
--- a/dt-eval-cli/README.md
+++ b/dt-eval-cli/README.md
@@ -342,7 +342,7 @@ dynatrace:
     apiToken: dt0s16.xxxxx          # needs storage:spans:read, storage:buckets:read
   destination:
     environmentUrl: https://eval-results.dev.apps.dynatracelabs.com
-    apiToken: dt0s16.yyyyy          # needs storage:bizevents:write, storage:metric:write
+    apiToken: dt0s16.yyyyy          # needs storage:events:write, storage:metrics:write
 ```
 
 Top-level `environmentUrl` / `apiToken` (if present) act as fallbacks — if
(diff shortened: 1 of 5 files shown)


yassine-dt (dismissed):
LGTM

PR #114, opened 2026-06-16
fix(dt-eval-lib): support AWS SSO/temporary creds and markdown-fenced JSON for Bedrock

armin-sams-dt:
## Background

A customer reported three bugs in the `bedrock` provider when using **AWS SSO / temporary credentials** (`AWS_SESSION_TOKEN`) and with models that wrap JSON in markdown code fences (`us.amazon.nova-2-lite-v1:0`). All three were verified against the code — the reported failures are real, though the *root-cause reasoning* for Bug 2 in the report was slightly off (see below).

## The 3 reported bugs & what changed

**Bug 1 — `sessionToken` not forwarded in the probe** (`dt-eval-cli/src/probe/provider.ts`)
The pre-flight check (`doctor`/`validate`) built `{ accessKeyId, secretAccessKey }` without `AWS_SESSION_TOKEN`, breaking SigV4 signing for temporary credentials (`The security token included in the request is invalid`).
→ Fixed.

**Bug 2 — `sessionToken` not forwarded during evaluation** (`dt-eval-lib/src/engine/providers/bedrock.ts`)
Same error on every `call()`. **Correction to the report's reasoning:** env-var users do *not* fall through to `credentials: undefined` — the factory (`providers/index.ts`) always resolved `AWS_ACCESS_KEY_ID`/`SECRET` from the environment and passed them on as `apiKey`/`secretKey`, so the static-credential branch always triggered and dropped `sessionToken`.
→ Fixed via the more robust **default-credential-chain approach**: explicit credentials are only built from configured static keys (incl. `sessionToken`); otherwise `credentials` is left undefined so the AWS SDK default credential chain resolves SSO, IAM roles, `AWS_PROFILE` and `AWS_SESSION_TOKEN` itself.

**Bug 3 — JSON wrapped in markdown fences** (`dt-eval-lib/src/engine/providers/bedrock.ts`)
`JSON.parse(text)` failed when the model wrapped its response in ` ```json … ``` `. Fixed with a new `stripCodeFences()` helper applied before `JSON.parse`.

**Additional (not in the report):** the eager `"Missing AWS credentials"` throw in the factory wrongly rejected pure IAM-role / `AWS_PROFILE` auth. Removed — credential resolution is now deferred to the SDK chain.

## Tests (TDD — red first, then green)

- `dt-eval-lib/tests/bedrock.test.ts` (new): sessionToken forwarding, `undefined` default-chain, fence stripping
- `dt-eval-cli/tests/probe/provider.test.ts`: 2 Bedrock probe tests added
- `dt-eval-lib/tests/engine.test.ts`: the test asserting the old exception updated to the new behavior

## Verification

- `dt-eval-lib`: lint clean, **124/124** tests passing, build ok
- `dt-eval-cli`: build/typecheck ok, all tests passing (incl. the 2 new probe tests)
- ⚠️ A real end-to-end run against Bedrock with SSO credentials was not performed (no AWS credentials available)

### DIFF
--- a/dt-eval-cli/src/probe/provider.ts
+++ b/dt-eval-cli/src/probe/provider.ts
@@ -121,18 +121,17 @@ export async function probeProvider(opts: ProbeOptions): Promise<ProbeResult> {
 
     if (provider === 'bedrock') {
       const region = opts.region ?? process.env['AWS_REGION'] ?? process.env['AWS_DEFAULT_REGION'] ?? 'us-east-1';
-      const sessionToken = process.env['AWS_SESSION_TOKEN'];
       // Lazy-import so non-Bedrock users don't pay the SDK init cost on every doctor/validate run.
       const { BedrockRuntimeClient, ConverseCommand } = await import('@aws-sdk/client-bedrock-runtime');
       // Only build an explicit credential object from configured static keys.
       // Otherwise leave credentials undefined so the AWS SDK default credential
-      // chain resolves them — this signs SSO / temporary env credentials
-      // (AWS_SESSION_TOKEN), IAM roles and AWS_PROFILE correctly. When static
-      // keys are configured, forward AWS_SESSION_TOKEN too.
+      // chain resolves them — this covers SSO / temporary env credentials
+      // (AWS_ACCESS_KEY_ID + AWS_SECRET_ACCESS_KEY + AWS_SESSION_TOKEN), IAM
+      // roles and AWS_PROFILE.
       const client = new BedrockRuntimeClient({
         region,
         credentials: opts.apiKey && opts.secretKey
-          ? { accessKeyId: opts.apiKey, secretAccessKey: opts.secretKey, ...(sessionToken ? { sessionToken } : {}) }
+          ? { accessKeyId: opts.apiKey, secretAccessKey: opts.secretKey }
           : undefined,
       });
       try {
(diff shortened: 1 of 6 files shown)


dt-eval-cli/src/probe/provider.ts
```diff
@@ -120,15 +120,19 @@ export async function probeProvider(opts: ProbeOptions): Promise<ProbeResult> {
...
+      // Otherwise leave credentials undefined so the AWS SDK default credential
+      // chain resolves them — this signs SSO / temporary env credentials
+      // (AWS_SESSION_TOKEN), IAM roles and AWS_PROFILE correctly. When static
+      // keys are configured, forward AWS_SESSION_TOKEN too.
       const client = new BedrockRuntimeClient({
         region,
-        credentials: accessKeyId && secretAccessKey
-          ? { accessKeyId, secretAccessKey }
+        credentials: opts.apiKey && opts.secretKey
+          ? { accessKeyId: opts.apiKey, secretAccessKey: opts.secretKey, ...(sessionToken ? { sessionToken } : {}) }
```
egehandulger: If `BedrockRuntimeClient` is able to pickup the env variables, do we need to catch the `AWS_SESSION_TOKEN` when we don't need to handle others?

yassine-dt:
image
Run with Bedrock session token works ✅ 
image
With Nova models ✅

dt-eval-cli/src/probe/provider.ts
yassine-dt: Good point I will fix that since I am doing end2end tests

egehandulger (approved):

krismuhi (approved):

PR #121, opened 2026-07-03
chore(dt-eval-lib): sync catalog-data.ts from internal engine

yassine-dt:
Auto-generated by `sync_prompts_to_catalog.py` from the internal eval-research-engine. Review the diff for prompt/scoring changes before merging.

**`npm test` FAILED** against the new catalog — this sync changes thresholds/requiredFields/scoring-type for some metrics, and `tests/prompts.test.ts` / `tests/engine.test.ts` still assert the old catalog's values. Update those tests before marking ready for review.

Opened as a draft — mark ready for review once verified.

Run manually from a local machine (interim, pending GitHub App cross-org credential — see `docs/prompt-catalog-sync-plan.md` in the internal repo).

### DIFF
--- a/dt-eval-lib/package.json
+++ b/dt-eval-lib/package.json
@@ -20,10 +20,7 @@
     "lint": "biome check src tests",
     "lint:fix": "biome check --write src tests",
     "format": "biome format --write src tests",
-    "format:check": "biome format src tests",
-    "e2e:azure": "tsx _temp_tests/azure-openai-e2e.ts",
-    "e2e:bedrock": "tsx _temp_tests/bedrock-e2e.ts",
-    "e2e:vertex": "tsx _temp_tests/vertex-e2e.ts"
+    "format:check": "biome format src tests"
   },
   "engines": {
     "node": ">=20"
--- a/dt-eval-lib/src/engine/providers/index.ts
+++ b/dt-eval-lib/src/engine/providers/index.ts
@@ -3,12 +3,12 @@ import type { ProviderOptions } from "../types";
 import type { LLMProvider } from "./types";
 
 const DEFAULT_MODELS: Record<string, string> = {
-  openai: "gpt-5.4",
-  anthropic: "claude-sonnet-4-6",
-  vertex: "gemini-3.1-pro-preview",
-  gemini: "gemini-3-flash-preview",
+  openai: "gpt-5-mini",
+  anthropic: "claude-haiku-4-5",
+  vertex: "gemini-3.1-flash-lite",
+  gemini: "gemini-3.1-flash-lite",
   // azure-openai: no default — deployment names are user-defined in Azure portal
-  bedrock: "us.anthropic.claude-3-5-haiku-20241022-v1:0",
+  bedrock: "global.anthropic.claude-haiku-4-5-20251001-v1:0",
 };
 
 const ENV_API_KEY: Record<string, string> = {
--- a/dt-eval-lib/src/prompts/catalog-data.ts
+++ b/dt-eval-lib/src/prompts/catalog-data.ts
@@ -2,402 +2,17 @@ import type { PromptDefinition } from "./types";
 
 export const catalog = [
   {
-    id: "fluency",
-    name: "Fluency",
-    version: "1.0.0",
-    description: "Evaluates whether the generated answer is well-written, coherent, and natural.",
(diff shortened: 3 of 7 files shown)


dt-eval-lib/package.json
```diff
@@ -20,10 +20,7 @@
     "lint": "biome check src tests",
     "lint:fix": "biome check --write src tests",
     "format": "biome format --write src tests",
-    "format:check": "biome format src tests",
-    "e2e:azure": "tsx _temp_tests/azure-openai-e2e.ts",
```
yassine-dt, line 24: Old slope

dt-eval-lib/src/engine/providers/index.ts
```diff
@@ -3,12 +3,12 @@ import type { ProviderOptions } from "../types";
 import type { LLMProvider } from "./types";
 
 const DEFAULT_MODELS: Record<string, string> = {
-  openai: "gpt-5.4",
-  anthropic: "claude-sonnet-4-6",
-  vertex: "gemini-3.1-pro-preview",
-  gemini: "gemini-3-flash-preview",
+  openai: "gpt-5-mini",
```
yassine-dt, line 6: @krismuhi  those are the models I used in the eval pipeline. I suggest we use the cheapest as defaults

krismuhi: perfect. we could call this out in the readme btw

dt-eval-lib/src/prompts/catalog-data.ts
```diff
@@ -474,35 +284,17 @@ score.value = 0.4×factual_correctness + 0.3×coverage + 0.1×conciseness + 0.2
...
... (\'shockingly\') signals exasperation.","signal":"sarcasm"}}\n\nMESSAGE: Homophobic AND racist. Yikes.\n{"score":0.0,"explanation":{"summary":"Sharp disapproval of described behaviour.","signal":"disapproval"}}\n\nMESSAGE: Would love to donate as well!\n{"score":1.0,"explanation":{"summary":"Positive and enthusiastic \u2014 no frustration signal.","signal":null}}\n\nMESSAGE: I\'m not sure I understand how to set this up, could you walk me through it?\n{"score":1.0,"explanation":{"summary":"Confusion and a polite request \u2014 no negative charge.","signal":null}}\n\nMESSAGE: The temperatures in Oymyakon must have slowed down that GIF.\n{"score":1.0,"explanation":{"summary":"Neutral factual or observational comment.","signal":null}}\n\nMESSAGE: My grandmother passed away last week and I still can\'t believe it.\n{"score":1.0,"explanation":{"summary":"Pure sadness without any object of complaint.","signal":null}}\n\nMESSAGE: I guess it is what it is.\n{"score":0.4,"explanation":{"summary":"Resigned acceptance with a hint of disappointment \u2014 mild frustration signal but could also be neutral acceptance; borderline.","signal":"resignation"}}\n\n<message>\n{{input}}\n</message>\n',
```
krismuhi, line 292: swear words are the most explicit way to collect user frustration especially with a chatapp

dt-eval-lib/src/prompts/catalog-data.ts
krismuhi, line 1: overall the dedicated examples feel like dont serve any purpose other than blowing up context window and send dumber models down the rabbithole. 

Can we pressure test this? 
What is we remove the theoretical examples? do we get better results across the datasets? 
for example: 
"ur solar system?\nOUTPUT: The eight planets are Mercury, Venus, Earth, Mars, Jupiter, Saturn, Uranus, and Neptune. Pluto was reclassified as a dwarf planet in 2006 and is no longer counted, though some still refer to it informally as a ninth planet.\n{"score":0.8,"explanation":"Core fact"

krismuhi (changes requested):

dt-eval-lib/src/engine/providers/index.ts
yassine-dt: sure my bad I will update the Readme

dt-eval-lib/src/prompts/catalog-data.ts
yassine-dt: I will drop then and compare. if the delta is small I will be happy to take the trade-off

yassine-dt: the context window is not an issue in my opinion. we will use slightly more tokens but  way cheaper models. When this can be an issue (which I didn't think of it untill now) is that the user don't like some of the examples

PR #122, opened 2026-07-03
fix(dt-eval-cli): stop sending evaluated prompt/response back to Dynatrace by default

krismuhi:
## Summary
- Bizevents written back to Dynatrace no longer include the evaluated prompt/response (`gen_ai.evaluation.input.question`, `.answer`, `.system_prompt`) unless explicitly opted in
- New `storeEvaluatedPrompt` YAML config field (default: `false`) and matching `--store-evaluated-prompt` CLI flag for customers who want that data stored in the result

## Test plan
- [x] `npx vitest run` — all 191 tests pass, including new coverage for the default-off / explicit-off / explicit-on cases
- [x] `npx tsc --noEmit` — no type errors
- [x] `npm run build` — builds cleanly

### DIFF
--- a/dt-eval-cli/src/cli/commands/run.ts
+++ b/dt-eval-cli/src/cli/commands/run.ts
@@ -81,6 +81,7 @@ export function createRunCommand(): Command {
   cmd.option('--dry-run', 'Fetch and transform traces, print payloads, do not send');
   cmd.option('--ci', 'Non-interactive mode: JSON stdout, exit 1 on threshold breach');
   cmd.option('--concurrency <n>', 'Number of parallel evaluation workers (overrides judge.concurrency in the config; default 5)', (v) => parseInt(v, 10));
+  cmd.option('--store-evaluated-prompt', 'Include the evaluated prompt/response in bizevents written to Dynatrace (overrides storeEvaluatedPrompt in the config; default: false)');
   cmd.option('--debug', 'Enable debug logging with per-step timing');
 
   cmd.action(async (configArg: string | undefined, options: {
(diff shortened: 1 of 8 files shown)


krismuhi:
tested on prod  UI renders properly. no issue 
image
image

dt-eval-cli/src/cli/commands/run.ts
```diff
@@ -81,6 +81,7 @@ export function createRunCommand(): Command {
   cmd.option('--dry-run', 'Fetch and transform traces, print payloads, do not send');
   cmd.option('--ci', 'Non-interactive mode: JSON stdout, exit 1 on threshold breach');
   cmd.option('--concurrency <n>', 'Number of parallel evaluation workers (overrides judge.concurrency in the config; default 5)', (v) => parseInt(v, 10));
+  cmd.option('--store-evaluated-prompt', 'Include the evaluated prompt/response in bizevents written to Dynatrace (overrides storeEvaluatedPrompt in the config; default: false)');
```
yassine-dt, line 84: Maybe we add the option of a negative flag to override it from the CLI
cmd.option('--store-evaluated-prompt', 'Include the evaluated prompt/response in bizevents (overrides storeEvaluatedPrompt in config; default: false)');

cmd.option('--no-store-evaluated-prompt', 'Exclude the evaluated prompt/response from bizevents (overrides storeEvaluatedPrompt: true in config)')

krismuhi: not sure the negated make sense, when the default is not to collect. or you are thinking as an additional layer, override even the YAML with this run command?

yassine-dt: yes it was a minor anyway

yassine-dt (approved):

PR #127, opened 2026-07-10
feat: support Vertex AI judge via Workload Identity / ADC

krismuhi:
## Summary

Adds `vertex` as a first-class judge provider that authenticates through Google **Application Default Credentials** (Workload Identity on GKE) rather than a static `GOOGLE_API_KEY`. This unblocks environments — like the customer (HCA) who reported this — whose security policy prohibits open API keys.

Previously, setting `judge.provider: vertex` failed config validation:

```
Config validation failed:
 - judge.provider must be one of: openai, anthropic, azure-openai, gemini, bedrock
```

## What changed

**dt-eval-lib**
- Split the `vertex` path from `gemini` in the provider factory. Vertex no longer falls back to `GOOGLE_API_KEY`; when no `apiKey` is supplied the `@google/genai` SDK resolves ADC automatically (`vertexai: true`). `project`/`location` resolve from options or `GOOGLE_CLOUD_PROJECT` / `GCLOUD_PROJECT` / `GCP_PROJECT_ID` and `GOOGLE_CLOUD_LOCATION` (default `us-central1`).
- `ProviderConfig.apiKey` is now optional; added `project`/`location` to `ProviderConfig` and `ProviderOptions`.

**dt-eval-cli**
- Added `vertex` to the config validation list and to `JudgeConfig` schema, with new `project`/`location` fields.
- Map GCP project/location env vars for `vertex` in config resolution (deliberately *not* auto-mapping `GOOGLE_API_KEY`).
- Added a `vertex` probe path so `doctor` / `validate` exercise ADC, threading `project`/`location` through `doctor`, `validate`, the probe, and the runner.

**Docs & tests**
- Documented Vertex/ADC setup in both READMEs.
- Updated the provider factory tests to reflect that Vertex no longer requires an API key.

## Origin

This reimplements a fix originally prototyped by HCA (support case TS022554579). Their fork was based on `0.2.5-alpha`, so it predated the Bedrock SSO/credential-chain fix (#114) and bundled internal-only CI. This PR rebuilds just the Vertex work on current `main`, leaving Bedrock untouched and dropping the internal artifacts.

## Testing

- `dt-eval-lib`: build + 123 tests pass.
- `dt-eval-cli`: build + 191 tests pass.
- Verified `dt-evals validate` now accepts `provider: vertex` and reaches the provider probe (which correctly requests project+location or a key when GCP credentials are absent) instead of rejecting the provider outright.


README.md
```diff
@@ -428,8 +428,11 @@ JUDGE_MODEL=gpt-4.1
 OPENAI_API_KEY=sk-...
 # Anthropic
 ANTHROPIC_API_KEY=sk-ant-...
-# Google (Vertex / Gemini)
+# Google Gemini Developer API
 GOOGLE_API_KEY=...
+# Google Vertex AI — uses Workload Identity / Application Default Credentials (no API key)
```
yassine-dt, line 433: I think  Vertex should be replaced with Gemini Enterprise Agent Platform  everywhere

dt-eval-lib/src/engine/providers/base.ts
```diff
@@ -1,7 +1,7 @@
 import type { LLMJudgeResponse, LLMProvider, ProviderConfig } from "./types";
 
 export abstract class BaseProvider implements LLMProvider {
-  protected readonly apiKey: string;
+  protected readonly apiKey: string | undefined;
```
yassine-dt: I think this will impact all other providers. A proper one might be override it only in Google but @egehandulger knows more here

yassine-dt:
the CLI wizard was not updated 
Screenshot 2026-07-10 at 13 25 57

I had to cerate a manual one and set vertex manually instead of gemini 
image

yassine-dt:
`run_id=run-2026-07-10T11-48-22-9d26c2b3`
The evals works using the ADC

yassine-dt:
Issues with the cli eval config 

- An old one: when you opt for default or ignore the api key it uses the open-ai key per default even if you choose a different provider 

image


I cerated an issue for this https://github.com/dynatrace-oss/dt-evals/issues/128

README.md
krismuhi: the good old rebranding struggle. they announced the blog and website
but when you login or check the DOCS, everything is still VERTEX. 
probably for search we need to use like Gemini Enterprise Agent Platform (Vertex AI)

yassine-dt:
Screenshot 2026-07-10 at 13 25 57

README.md
yassine-dt: okey all good

dt-eval-lib/src/engine/providers/base.ts
egehandulger: If it's breaking other providers when it's empty, then a runtime check would be necessary.

krismuhi: Good catch — fixed in 678de0b by reverting the type change entirely rather than adding a runtime check.

`BaseProvider.apiKey` and `ProviderConfig.apiKey` are back to the required `string` type, so no other provider is affected. Vertex now follows the exact idiom Bedrock already uses for keyless auth (`index.ts`): the factory passes `options.apiKey ?? ""`, and `GoogleProvider` treats an empty string as "no key" (`...(config.apiKey ? { apiKey } : {})`), letting the `@google/genai` SDK resolve ADC / Workload Identity. `base.ts` now has zero diff against `main`.

This keeps the strictness @egehandulger wanted without scattering runtime checks across providers, since `createProvider` is the single construction path and already guards the key-requiring providers.

krismuhi: addressed. good callout

dt-eval-cli/src/config/defaults.ts
```diff
@@ -50,7 +50,7 @@ export const DEFAULT_CONFIG: Omit<DtEvalConfig, 'dynatrace' | 'judge'> & {
 export const DEFAULT_JUDGE_MODELS: Record<string, string> = {
   openai: 'gpt-4.1',
   anthropic: 'claude-sonnet-4-6',
-  vertex: 'gemini-2.5-pro',
+  vertex: 'gemini-3.5-flash',
   gemini: 'gemini-2.5-flash',
   // azure-openai: no default — deployment names are user-defined in Azure portal
   bedrock: 'us.anthropic.claude-3-5-haiku-20241022-v1:0',
```
thisthat, line 56: we should setup an automation to keep track of new releases, similar to Renovate

dt-eval-cli/src/probe/provider.ts
```diff
@@ -149,6 +153,36 @@ export async function probeProvider(opts: ProbeOptions): Promise<ProbeResult> {
...
+      // Vertex AI probe — uses @google/genai with vertexai:true.
+      // API key is optional: GKE Workload Identity / ADC will be used automatically when no key is set.
+      // Do not auto-map GOOGLE_API_KEY env here; default probe path should validate ADC/WI.
+      const apiKey = opts.apiKey;
+      const project =
+        opts.project ??
+        process.env['GOOGLE_CLOUD_PROJECT'] ??
+        process.env['GCLOUD_PROJECT'] ??
+        process.env['GCP_PROJECT_ID'];
+      const location = opts.location ?? process.env['GOOGLE_CLOUD_LOCATION'] ?? 'us-central1';
```
thisthat: Here we default to `us-central-1`, but [here](https://github.com/dynatrace-oss/dt-evals/pull/127#discussion_r3576898560) we default to `global`

dt-eval-cli/src/cli/commands/configure.ts
```diff
@@ -267,6 +277,23 @@ export function createConfigureCommand(): Command {
...
+          default:
+            existing.judge?.project ??
+            process.env['GOOGLE_CLOUD_PROJECT'] ??
+            process.env['GCLOUD_PROJECT'] ??
+            process.env['GCP_PROJECT_ID'] ??
+            '',
+        });
+        vertexLocation = await input({
+          message: 'Vertex AI location (for example global or us-central1)',
+          default: existing.judge?.location ?? process.env['GOOGLE_CLOUD_LOCATION'] ?? 'global',
```
thisthat, line 295: Here we default to `global`, but [here](https://github.com/dynatrace-oss/dt-evals/pull/127#discussion_r3576896133) we default to `us-central-1`

dt-eval-cli/src/config/defaults.ts
yassine-dt: for this @thisthat once I migrate to the new repo of evals. I will set up automation for default prompts and defaults models that we test so we are sure they are tested and work well with our prompts

dt-eval-cli/src/probe/provider.ts
krismuhi: Fixed in 0101089 — standardized on `global` everywhere. Since the default Vertex model is now `gemini-3.5-flash` (served via the Vertex global endpoint), `global` is the correct default; `us-central1` here was the stale side. The lib factory, probe, schema docs, and both READMEs now all default to `global`, matching `configure`.

dt-eval-cli/src/cli/commands/configure.ts
krismuhi: `global` here is correct (it pairs with the `gemini-3.5-flash` default model, which is served via the Vertex global endpoint). The inconsistency was on the other side — probe/factory/docs still said `us-central1`. Aligned all of them to `global` in 0101089, so configure/validate/doctor/run now resolve the same location.

PR #129, opened 2026-07-10
fix(dt-eval-cli): avoid leaking api keys across providers

yassine-dt:
## Problem
`dt-evals configure` writes credentials into the generated config and can also carry a stale API key across provider switches.

## Solution
Write only non-secret settings to YAML, persist local secrets via the existing `.env` flow, and stop reusing a previous provider key when the provider changes.


yassine-dt:
Before
image
After 
image

krismuhi (approved):

PR #135, opened 2026-07-20
fix(dt-eval-cli): default structured input to last user message

yassine-dt:
## Summary
- default structured `gen_ai.input.messages` to the last extracted user message
- preserve explicit `scope.spanFields.input` overrides and plain-string input behavior
- cover multipart user messages and updated parser expectations in tests

## Testing
- npm test


krismuhi (approved):
looks good. great catch

PR #136, opened 2026-07-21
feat(dt-eval-cli): separate evaluator context from systemInstruction

yassine-dt:
Summary
- add first-class scope.spanFields.context
- add canonical context for per-metric routing
- default evaluator context to canonical context instead of systemInstruction
- keep systemInstruction reserved for actual system prompts

Behavior
Defaults
- input still defaults from gen_ai.input.messages
- output keeps the existing output parsing behavior
- context has no built-in default span attribute, so evaluator context is omitted unless configured
- systemInstruction still resolves from system-prompt sources, but it is no longer used as evaluator context by default

Custom mapping
- scope.spanFields.context maps a span attribute into canonical context
- scope.spanFields.systemInstruction maps a span attribute into canonical systemInstruction
- plain/default metric execution now uses span.context for evaluator context
- metrics can still explicitly route inputs.context: systemInstruction when they really want the system prompt

Fallbacks
- when a metric routes inputs.context to another canonical field and that field is empty, the runner falls back to canonical context
- if canonical context is also empty, evaluator context is omitted

Testing
- npm test

### DIFF
--- a/dt-eval-cli/README.md
+++ b/dt-eval-cli/README.md
@@ -397,6 +397,7 @@ scope:
   spanFields:
     input: llm.user_input              # or [llm.user_input, my.custom.input]
     output: llm.response
+    context: rag.documents
     systemInstruction: llm.system
     model: llm.model
 ```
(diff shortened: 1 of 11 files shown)


dt-eval-cli/README.md
```diff
@@ -434,14 +435,16 @@ metrics:
         input: userPrompt                           # latest user-role prompt slot
     - id: hallucination
       inputs:
-        context: systemInstruction                  # use system prompt as context
+        context: context                            # use the canonical context field for this metric
 ```
 
-Available canonical fields: `input`, `output`, `systemInstruction`, `model`,
-`userPrompt` (latest user-role prompt slot, extracted from
+Available canonical fields: `input`, `output`, `context`,
```
krismuhi, line 441: weird lionebreaks

krismuhi (approved):

PR #138, opened 2026-07-22
fix(dt-eval-cli): suggest correct default model per provider in wizard

armin-sams-dt:
## Problem (AI-315)

The `configure` wizard always suggested **and persisted** the OpenAI default model (`gpt-5-mini`) regardless of the selected provider.

**Root cause:** `resolveEffectiveConfig` injects `judge.model = DEFAULT_JUDGE_MODELS[provider]` for the default provider (`openai`), so `loadConfig()` returned `gpt-5-mini` even for a fresh setup. Every model code path in `configure` then read `existing.judge?.model` blindly.

## Fix

Added `suggestModelForProvider()` (in `config/defaults.ts`) which reuses an existing model **only when the provider is unchanged**, otherwise returns that provider's default — or `''` for `azure-openai`, whose deployment names are user-defined. Wired into all three leaking paths:

1. interactive model-prompt default
2. interactive save fallback
3. flag-only mode save (`configure --provider …`)

## Result (per provider)

| Provider | Model |
|---|---|
| openai | `gpt-5-mini` |
| anthropic | `claude-sonnet-4-6` |
| gemini / vertex | `gemini-3.1-flash-lite` |
| bedrock | `us.anthropic.claude-haiku-4-5-20251001-v1:0` |
| azure-openai | *(unset — deployment name required)* |

## Testing

- Unit tests for `suggestModelForProvider` (`tests/config-defaults.test.ts`).
- Integration test driving the real CLI flag-mode end-to-end (`tests/configure.integration.test.ts`).
- Manually verified the interactive wizard and flag mode for all six providers.
- Full suite: 211/211 green; build + typecheck green.

### DIFF
--- a/dt-eval-cli/tests/config-defaults.test.ts
+++ b/dt-eval-cli/tests/config-defaults.test.ts
@@ -1,0 +1,60 @@
+import { describe, it, expect } from 'vitest';
+import { suggestModelForProvider, DEFAULT_JUDGE_MODELS } from '../src/config/defaults.js';
+import type { DtEvalConfig } from '../src/config/schema.js';
+
+type Provider = DtEvalConfig['judge']['provider'];
+
+// Derive everything from the single source of truth so updating a model name in
+// defaults.ts never breaks these tests — they assert the routing logic, not the
+// literal model strings.
+const providersWithDefault = Object.keys(DEFAULT_JUDGE_MODELS) as Provider[];
+
+describe('suggestModelForProvider', () => {
+  it('suggests each provider default from DEFAULT_JUDGE_MODELS when no existing config', () => {
+    expect(providersWithDefault.length).toBeGreaterThan(0);
+    for (const provider of providersWithDefault) {
+      expect(suggestModelForProvider(provider)).toBe(DEFAULT_JUDGE_MODELS[provider]);
+    }
+  });
+
+  // Regression for AI-315: switching provider must not carry over the previous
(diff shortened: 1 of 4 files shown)


dt-eval-cli/tests/config-defaults.test.ts
```diff
@@ -0,0 +1,43 @@
+import { describe, it, expect } from 'vitest';
+import { suggestModelForProvider, DEFAULT_JUDGE_MODELS } from '../src/config/defaults.js';
+
+describe('suggestModelForProvider', () => {
+  it('suggests the provider-specific default when no existing config', () => {
+    expect(suggestModelForProvider('anthropic')).toBe('claude-sonnet-4-6');
```
yassine-dt: Can we avoid hard-coding current default model names in `config-defaults.test.ts` and derive them from `DEFAULT_JUDGE_MODELS` instead? Right now the test will break on intentional default-model updates. For the provider-switch regression, using something like `custom-openai-model` would keep the assertion focused on “ignore previous provider’s model” rather than OpenAI’s current default.

yassine-dt (changes requested):

armin-sams-dt (commented):
review finished.
also changed the tests to properly fetch the models from defaults.ts

yassine-dt (approved):
LGTM

krismuhi (approved):

PR #143, opened 2026-07-22
docs(main): correct bizevent read scope (AI-319)

yassine-dt:
## Summary
- document `storage:bizevents:read` as the drift baseline scope used by `fetch bizevents`
- remove root README optional scopes that are not part of the current CLI token guidance
- clarify the destination token requirements in the cross-tenant example

## Testing
- docs only


yassine-dt:
Reverting this PR because Dynatrace runtime currently returns missing scope `storage:bizevents:read` for `fetch bizevents`, so the scope-name change was incorrect.

krismuhi (approved):

PR #147, opened 2026-07-23
fix(dt-eval-cli): remove batch head-of-line blocking in eval runner

### LINKED ISSUE #152: Prevent eval runner slowdowns from batch head-of-line blocking
## Problem

The eval runner loses throughput because its batching logic processes work in fixed groups and waits for the whole group to finish before starting the next one.

With concurrency greater than 1, one slow evaluation can leave other worker slots idle after faster items complete. That increases total runtime and makes CI timeouts more likely under larger span x metric workloads.

yassine-dt:
Closes #152

## Summary

This PR only changes the eval runner batching logic and its tests.

## Code changes

- `dt-eval-cli/src/runner/batch.ts`
  - replace fixed-group batching with a work-conserving worker pool
  - preserve result ordering while starting new work as soon as a worker becomes free
  - reject invalid `concurrency` values
- `dt-eval-cli/tests/runner/batch.test.ts`
  - add coverage for work-conserving dispatch
  - add coverage for invalid concurrency

## Why this helps

The previous implementation waited for a whole batch group to finish before starting the next one. That created idle worker slots whenever one item in the group was slower than the others. The new worker-pool implementation keeps available slots busy until the queue is drained, which improves throughput without changing eval behavior.

## Validation

- targeted runner tests
- CLI build

### DIFF
--- a/dt-eval-cli/src/runner/batch.ts
+++ b/dt-eval-cli/src/runner/batch.ts
@@ -17,7 +17,7 @@ export async function processBatch<T, R>(
   handler: (item: T) => Promise<R>,
   config?: Partial<BatchConfig>,
 ): Promise<Array<BatchItemResult<T, R>>> {
-  const { concurrency } = { ...DEFAULT_BATCH_CONFIG, ...config };
+  const concurrency = config?.concurrency ?? DEFAULT_BATCH_CONFIG.concurrency;
   if (!Number.isInteger(concurrency) || concurrency <= 0) {
     throw new Error(`concurrency must be a positive integer, got ${concurrency}`);
   }
--- a/dt-eval-cli/src/runner/batch.ts
+++ b/dt-eval-cli/src/runner/batch.ts
@@ -1,6 +1,5 @@
 export interface BatchConfig {
   concurrency: number; // default 5
-  batchSize: number;   // default 10
 }
 
 export interface BatchItemResult<T, R> {
@@ -11,7 +10,6 @@ export interface BatchItemResult<T, R> {
 
 const DEFAULT_BATCH_CONFIG: BatchConfig = {
   concurrency: 5,
-  batchSize: 10,
 };
 
 export async function processBatch<T, R>(
--- a/dt-eval-cli/src/runner/batch.ts
+++ b/dt-eval-cli/src/runner/batch.ts
@@ -20,32 +20,29 @@ export async function processBatch<T, R>(
   config?: Partial<BatchConfig>,
 ): Promise<Array<BatchItemResult<T, R>>> {
   const { concurrency } = { ...DEFAULT_BATCH_CONFIG, ...config };
-  const results: Array<BatchItemResult<T, R>> = [];
+  if (!Number.isInteger(concurrency) || concurrency <= 0) {
+    throw new Error(`concurrency must be a positive integer, got ${concurrency}`);
+  }
 
-  // Process with concurrency limit using a semaphore-like approach
-  const queue = [...items];
-  const inFlight: Promise<void>[] = [];
+  const results = new Array<BatchItemResult<T, R>>(items.length);
+  let nextIndex = 0;
 
-  async function processItem(item: T): Promise<void> {
-    try {
-      const result = await handler(item);
-      results.push({ item, result });
-    } catch (err) {
-      results.push({
-        item,
-        error: err instanceof Error ? err : new Error(String(err)),
-      });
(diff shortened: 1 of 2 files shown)


dt-eval-cli/src/runner/batch.ts
armin-sams-dt: do we still need that value?

armin-sams-dt: do we still need that value?

armin-sams-dt (approved):

armin-sams-dt (changes requested):
Added 2 suggestions

dt-eval-cli/src/runner/batch.ts
yassine-dt: good catch I delete it

yassine-dt: yes I fixed it

dt-eval-cli/src/runner/batch.ts
```diff
@@ -20,32 +18,29 @@ export async function processBatch<T, R>(
   config?: Partial<BatchConfig>,
 ): Promise<Array<BatchItemResult<T, R>>> {
   const { concurrency } = { ...DEFAULT_BATCH_CONFIG, ...config };
```
armin-sams-dt: Nit (low severity): `{ ...DEFAULT_BATCH_CONFIG, ...config }` copies an explicit `undefined` over the default. If a caller passes `{ concurrency: undefined }`, the spread overwrites `5` with `undefined`, which then trips the guard below and throws — even though the `Partial<BatchConfig>` signature suggests the field is optional and should fall back to the default.

Suggestion:
```suggestion
  const concurrency = config?.concurrency ?? DEFAULT_BATCH_CONFIG.concurrency;
```
`??` falls back to the default for both a missing and an explicitly-`undefined` field, while real bad values (`0`, negative, `NaN`) still hit the guard. Not live today (the only caller passes `opts.concurrency ?? 5`), so purely a latent robustness fix.

dt-eval-cli/tests/runner/batch.test.ts
```diff
@@ -79,6 +79,30 @@ describe('processBatch', () => {
     expect(maxConcurrent).toBeLessThanOrEqual(3);
```
armin-sams-dt, line 79: Nit (low severity): this only asserts an upper bound, so it guards against *too much* parallelism but not *too little*. A pool that regressed to running serially (or with only 1 worker) would still satisfy `<= 3` and pass green — even though under-utilization is exactly the regression this PR is meant to prevent.

Suggestion — assert the limit is actually reached (20 items with equal-length `setTimeout` handlers saturate the 3 slots deterministically):
```suggestion
    expect(maxConcurrent).toBe(3);
```
If `toBe(3)` feels too strict, keep the existing line and add `expect(maxConcurrent).toBeGreaterThanOrEqual(3);` — that covers the lower bound while staying tolerant.

dt-eval-cli/tests/runner/batch.test.ts
```diff
@@ -107,4 +131,9 @@ describe('processBatch', () => {
     await processBatch(items, handler);
     expect(maxConcurrent).toBeLessThanOrEqual(5);
   });
+
+  it('rejects invalid concurrency', async () => {
+    await expect(processBatch([1], async n => n, { concurrency: 0 }))
+      .rejects.toThrow('concurrency must be a positive integer');
+  });
```
armin-sams-dt: Nit (low severity): this covers only `0`, but the guard (`!Number.isInteger(concurrency) || concurrency <= 0`) also rejects negatives, non-integers and `NaN`. `NaN` is the meaningful one: a bad `--concurrency` CLI arg becomes `parseInt(...) -> NaN`, which the *old* group-based code silently swallowed into an empty result set (a run that quietly does nothing). This new guard turns that into a clear error — worth locking in with a test.

Suggestion:
```suggestion
  it.each([0, -1, 2.5, NaN])(
    "rejects invalid concurrency: %s",
    async invalid => {
      await expect(processBatch([1], async n => n, { concurrency: invalid }))
        .rejects.toThrow("concurrency must be a positive integer");
    },
  );
```

dt-eval-cli/tests/runner/batch.test.ts
yassine-dt: `We respects concurrency limit by not exceeding it ` we are testing to exceed the limit so it should be less than or equal

armin-sams-dt (approved):

PR #154, opened 2026-07-23
fix: show evaluator result ratios

yassine-dt:
## Summary
- report per-evaluator results as success/total ratios instead of redistributed aggregate counts
- report errors as error/total ratios per evaluator
- relabel evaluator timing to Avg Duration and compute it from evaluator calls
- keep persisted run records unchanged

## Validation
- npm test
- npm run build

### DIFF
--- a/dt-eval-cli/src/cli/commands/run.ts
+++ b/dt-eval-cli/src/cli/commands/run.ts
@@ -73,12 +73,11 @@ export function buildEvaluationResultRows(metrics: string[], result: RunResult):
     const metricResult = byMetric.get(m);
     const total = metricResult?.total ?? 0;
     const successes = metricResult?.successes ?? 0;
-    const errors = metricResult?.errors ?? 0;
+    const passPercentage = total === 0 ? 0 : Math.round((successes / total) * 100);
 
     return [
       m,
-      `${successes}/${total}`,
-      `${errors}/${total}`,
+      `${successes}/${total} (${passPercentage}% passed)`,
       formatDuration(metricResult?.avgDurationMs ?? 0),
     ];
   });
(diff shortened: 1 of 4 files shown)


dt-eval-cli/src/runner/index.ts
```diff
@@ -244,6 +279,12 @@ export async function runEvals(
         evalCount++;
         evalErrors++;
         const elapsed = Date.now() - t0;
+        const stats = evaluatorStats.get(task.metric);
+        if (stats) {
+          stats.total++;
+          stats.errors++;
+          stats.durationMs += elapsed;
+        }
```
thisthat: this piece of code could benefit from some refactoring to make it easier to human read 🤔 
not a blocker for this PR tho

dt-eval-cli/src/cli/commands/run.ts
```diff
@@ -67,6 +67,23 @@ function evalErrorHint(messages: string[]): string | null {
...
+  return metrics.map(m => {
+    const metricResult = byMetric.get(m);
+    const total = metricResult?.total ?? 0;
+    const successes = metricResult?.successes ?? 0;
+    const errors = metricResult?.errors ?? 0;
+
+    return [
+      m,
+      `${successes}/${total}`,
+      `${errors}/${total}`,
```
thisthat: I wonder if this is easy to read 🤔 Maybe we should report % and only successes, eg `37/90 (41% passed)` 🤔

yassine-dt:
The new view
image

dt-eval-cli/src/cli/commands/run.ts
yassine-dt: Good catch will do

dt-eval-cli/src/runner/index.ts
yassine-dt: agreed

thisthat (approved):

PR #159, opened 2026-07-24
fix(dt-eval-cli): ai-336 align output messages parsing

### LINKED ISSUE #157: Align output parsing with gen_ai.output.messages semconv
## Problem

The evaluator output mapping currently still prioritizes / supports the legacy singular span attribute `gen_ai.output.message`. Current OTel GenAI semantic conventions use `gen_ai.output.messages` (plural) for model output choices.

This can cause structured assistant messages, especially tool-call responses, to be sent to evaluators or stored in evaluation bizevents as raw JSON instead of being parsed into the evaluated assistant output.

## Desired behavior

- Remove `gen_ai.output.message` from the default output attribute mapping.
- Treat `gen_ai.output.messages` as the canonical default output source.
## Acceptance criteria

- Default output candidates no longer include `gen_ai.output.message`.

yassine-dt:
## Summary

- remove non-semconv `gen_ai.output.message` from default output candidates
- keep `gen_ai.output.messages` as the default output source
- keep legacy singular output support available only through explicit `scope.spanFields.output`

## Validation

- `npm test`
- `npm run build`

Closes #157.

### DIFF
--- a/dt-eval-cli/README.md
+++ b/dt-eval-cli/README.md
@@ -385,14 +385,13 @@ for end-to-end instrumentation examples.
 Each entry accepts a single attribute or a list of candidates; the first
 non-null value wins, with the built-in defaults appended as fallback.
 
-**Example 1 — explicit legacy output mapping** (only needed for emitters
-that still use the non-semconv singular attribute `gen_ai.output.message`):
+**Example 1 — explicit legacy output mapping**:
 
 ```yaml
 scope:
   service: pydantic-ai-music-agent
   spanFields:
-    output: gen_ai.output.message
+    output: `gen_ai.completion.0.content
 ```
 
 The default output mapping uses `gen_ai.output.messages`. `spanFields` lets
--- a/dt-eval-cli/README.md
+++ b/dt-eval-cli/README.md
@@ -374,7 +374,7 @@ For `dt-evals alerts apply` (Dynatrace Workflows), grant these additionally on t
 ### Custom span field mapping
 
 By default, the CLI reads OTel GenAI semconv (`gen_ai.input.messages`,
-`gen_ai.output.messages`, `gen_ai.output.message`, …). It also probes a
+`gen_ai.output.messages`, …). It also probes a
 small set of legacy fallback attributes (`gen_ai.prompt.N.*`,
 `gen_ai.completion.0.content`) for compatibility with older emitters. If
 your spans expose the LLM I/O under different attribute names — for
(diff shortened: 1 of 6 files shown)


dt-eval-cli/tests/dt/dql.test.ts
```diff
@@ -69,7 +69,8 @@ describe('buildGenAiSpanQuery', () => {
     const query = buildGenAiSpanQuery({ since: '1h' });
     // OTel GenAI fields
     expect(query).toContain('gen_ai.input.messages');
-    expect(query).toContain('gen_ai.output.message');
+    expect(query).toContain('gen_ai.output.messages');
+    expect(query).not.toContain('gen_ai.output.message,');
```
armin-sams-dt, line 73: is the check for `NOT .message` really necessary?

yassine-dt: to make sure we don't use `gen_ai.output.message` again

armin-sams-dt:
Closes #157 isn't fully met: 
- parsing of tool_call parts in gen_ai.output.messages is missing.

1.  extractRolesFromJsonMessages (dql.ts) only reads `content/text`parts -> a tool_call part has neither (instead has: `name` and `arguments`) and thus gets skipped
  - tool_call-only output → nothing extracted → span silently dropped at if (!input || !output) continue.
  - text + tool_call mixed → only the text survives, tool_call is lost.

  Also, no test covers this. Needs a tool_call branch + tests for tool-call-only and mixed.
  
 maybe remove that from the issue and create a new issue, so that the messages/message problem is solved in this PR and the other one in a another PR ?

dt-eval-cli/tests/dt/dql.test.ts
armin-sams-dt: Okay, makes sense. However:

It only catches the comma variant, if gen_ai.output_message ever ends up as the last field, and hence has no trailing comma, it wouldn't be caught.

```suggestion
    expect(query).not.toMatch(/gen_ai\.output\.message\b(?!s)/);
```

this should then match both `gen_ai.output.message` and `gen_ai.output.message,`

dt-eval-cli/README.md
```diff
@@ -385,22 +385,18 @@ for end-to-end instrumentation examples.
...
-`gen_ai.output.message`):
+**Example 1 — explicit legacy output mapping** (only needed for emitters
+that still use the non-semconv singular attribute `gen_ai.output.message`):
 
 ```yaml
 scope:
   service: pydantic-ai-music-agent
   spanFields:
-    output: [gen_ai.output.message, gen_ai.output.messages]
+    output: gen_ai.output.message
```
armin-sams-dt: this example shows the legacy output. 
However:
`vals/pydantic-ai-music-agent.dt-eval.yaml:23`,
as well as
`evals/pydantic-ai-music-agent.dt-eval.yaml:23` 
use `gen_ai.output.messages`.
So there's a mismatch. Please check if thats fine or not.

armin-sams-dt (changes requested):
Added some suggestions.
The mismatch in the legacy example is still open.

dt-eval-cli/README.md
yassine-dt: yes fixed that thanks

thisthat (approved):

dt-eval-cli/tests/dt/dql.test.ts
thisthat: Good catch 👍

armin-sams-dt (approved):

PR #171, opened 2026-07-30
fix(dt-eval-cli): route ingest to env-api on labs tenants

armin-sams-dt:
## Why

Users on internal labs/sprint tenants (`*.apps.dynatracelabs.com`) can't run an eval end-to-end: the BizEvent ingest fails with `400 "Invalid app context"`, even with a platform token that has every scope. Spans get fetched and tasks get judged, then the run dies right before writing results. A production customer hit the same wall.

The token was never the problem. The request was going to the wrong host.

## What was wrong

Ingest posts to the classic environment API at `/api/v2/bizevents/ingest`. DQL queries go to the platform host (`*.apps.*`). Since users configure a single `.apps.` URL, the client rewrites that host to the env-api host for ingest only — that's what `liveSubdomain()` does.

That rewrite only knew the production rule:

```
{env}.apps.dynatrace.com  ->  {env}.live.dynatrace.com
```

For a labs host like `uim8926h.sprint.apps.dynatracelabs.com` it matched nothing and left the URL alone, so ingest hit the **apps gateway** instead of the env-api. The gateway rejects that with `400 "Invalid app context"`. DQL kept working because it's supposed to stay on `.apps.`.

Labs env-api hosts don't become `.live.` — they just drop the `apps.` label. This is confirmed by a working manual ingest and by the fact that `dtctl/index.ts` already carried this exact handling; only the copy in `dt/client.ts` was missed.

```
{env}.sprint.apps.dynatracelabs.com  ->  {env}.sprint.dynatracelabs.com
```

## How

- **`dt/client.ts`:** add the labs rule to `liveSubdomain()` as a second, non-overlapping replace (the production regex requires the literal `dynatrace.com`, so it can't match `dynatracelabs.com`). Refresh the now-inaccurate `.live.` doc and inline comments.
- **`dtctl/index.ts`:** fix the scope label `doctor` shows labs users. It reported `openpipeline:bizevents:ingest`, but the probe is redirected to the classic env-api, which wants `storage:events:write`. `isClassic` now derives from whether `liveSubdomain()` rewrites the host, so the label always matches where the request actually goes.
- **`dtctl/index.ts` (`createPlatformToken`):** it built `/api/v2/apiTokens` (a classic env-api endpoint) from the raw URL, so on any `.apps.` host it would hit the apps gateway too. Routed through `liveSubdomain()` for consistency. This function is exported but currently unused, so it's defensive — no live path hit it, but it was the same footgun waiting to happen.
- **Tests:** regression cases for the sprint and dev labs variants, a check that DQL stays on `.apps.` for labs, and a URL-shape test for `createPlatformToken`.

## Scope check

Traced the write path to be sure the fix actually reaches it. The cross-tenant `destination` tenant (README's documented labs example, `eval-results.dev.apps.dynatracelabs.com`) is a plain `DynatraceClient`, so its ingest goes through the same rewrite — no separate code path. DQL builders correctly stay on `.apps.` and were left alone. `createPlatformToken` was the only other classic env-api URL missing the rewrite.

## Testing

- Full CLI suite green: 206 passed.
- Typecheck clean, build succeeds.

## Auth scheme — verified

The original `400` was a routing error that fired *before* auth, so it left one thing unproven: this client sends `Bearer dt0s16…` for platform tokens, while the reported working curl used `Api-Token dt0s16…`. I verified our scheme by posting a single probe bizevent through the fixed client to a production `.apps.dynatrace.com` tenant with a `dt0s16` platform token: `Bearer` → classic env-api `/api/v2/bizevents/ingest` returns 2xx. So the two halves are both confirmed — the labs *host* by the working curl, the `Bearer` *auth* by this probe. There's no reason auth would differ by env-api host family.

The one combination not yet exercised on a live tenant is `Bearer` + a labs `.dynatracelabs.com` env-api at the same time (I don't have a labs token). A post-merge run on the labs tenant is a cheap final confirmation; a `401`/`403` there would be a separate auth issue, not a regression of this fix.

### DIFF
--- a/dt-eval-cli/src/dt/client.ts
+++ b/dt-eval-cli/src/dt/client.ts
@@ -70,22 +70,30 @@ const POLL_INTERVAL_MS = 1000;
 const POLL_TIMEOUT_MS = 60_000;
 
 /**
- * Translate `*.apps.dynatrace.com` → `*.live.dynatrace.com`.
+ * Rewrite a platform (`.apps.`) host to its classic environment-API host.
  *
  * Modern Dynatrace platform splits its surface across two subdomains:
  *   - `.apps.` hosts the platform-storage / DQL APIs (`/platform/storage/...`).
- *   - `.live.` hosts the classic environment APIs (`/api/v2/...`).
+ *   - the env-api host hosts the classic environment APIs (`/api/v2/...`).
  *
  * Users supply a single `environmentUrl` for the tenant — by convention the
  * `.apps.` form (because DQL is the entry point). For ingest endpoints
  * (`/api/v2/bizevents/ingest`, `/api/v2/metrics/ingest`) we need to redirect
- * to `.live.`, otherwise the apps gateway returns 400 *Invalid app context*.
+ * to the env-api host, otherwise the apps gateway returns 400 *Invalid app
+ * context*.
  *
- * Idempotent on URLs that already point at `.live.` or any other host
+ * The env-api host differs by environment family:
+ *   - Production SaaS: `{env}.apps.dynatrace.com` → `{env}.live.dynatrace.com`.
+ *   - Internal labs (sprint/dev): the env-api simply drops the `apps.` label,
+ *     e.g. `{env}.sprint.apps.dynatracelabs.com` → `{env}.sprint.dynatracelabs.com`.
+ *
+ * Idempotent on URLs that already point at the env-api host or any other host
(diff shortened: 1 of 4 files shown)


yassine-dt (approved):
LGTM but please check manually if both the prod tenant and sprint tenants works with this

thisthat (approved):
Good catch 👍

armin-sams-dt (commented):
LGTM

PR #177, opened 2026-08-06
test: add end-to-end suite for the dt-evals cli

armin-sams-dt:
## Why

Until now, dt-evals only had unit tests. This PR adds an end-to-end suite: it runs the real, built CLI against a real Dynatrace tenant and a real judge, then checks what comes back. Nothing in the eval path is mocked — the CLI runs as a subprocess, the same way a user would run it.

Design: AI-309. Implementation: AI-310.

## What's in it

Seven suites, each covering one thing:

- **Harness logic** — the suite's own gating (skip vs. fail) and secret-redaction code, tested on its own.
- **Test rig** — the CLI runs as an isolated subprocess (its own `HOME`, its own working directory), with nothing leaking in from the developer's machine.
- **Tenant reachability** — a valid token connects, a bad one is rejected.
- **Fixture contract** — every span attribute dt-evals depends on, checked against the real fixture data.
- **`validate`** — every pre-flight check, both passing and failing.
- **`run`** — every run mode, including the full round trip into the tenant.
- **`evallogic`** — whether the evaluation is actually *correct*, not just whether data moved.

It lives in its own workspace, not inside the CLI's own tests, so judge spend and real secrets can never leak onto the pull-request path.

CI runs the full suite weekly, in two lanes. Real breakage — auth, connectivity, a broken fixture contract — blocks the run. A judge disagreeing about a verdict only alerts, since that's expected noise, not a hard failure. A separate, credential-free check runs on every PR instead. Failures land in one GitHub issue that closes itself once things recover.

## Follow up refinement

Four small product gaps remain in dt-evals, all tracked under epic AI-298. None of them block this suite — it already runs and passes. dynatrace-oss/dt-evals#179 fixes all four:

- **AI-387 — filter spans by attribute.** The costliest gap: without it, `evallogic` scores about 40 spans just to check 2. It's also what's blocking scoping the other 13 evaluators the way the design doc wants.
- **AI-388 — group multi-turn conversations.** Each fixture turn is its own trace today, so a multi-turn case never groups into one conversation. Not a rollback of AI-299 — the open question is whether we should evaluate a conversation once, or once per turn.
- **AI-389 — sort before the row limit.** `sampling: latest` isn't reliably "latest" once a service has more spans than the row limit.
- **AI-390 — tag bizevents with custom attributes.** There's no way to tag test records today, so nothing can find and clean them up later.

## Verification

**Automated:** typecheck is clean. 55 tests pass, 31 of them without any credentials. `E2E_REQUIRE_ENV=1` correctly turns a missing credential into a failure instead of a silent skip.

**Not yet run:** the suite's own test files — `validate.e2e.test.ts`, `run.e2e.test.ts`, `evallogic.e2e.test.ts` — have never executed against a real tenant. They're typechecked and reviewed against the CLI's source, nothing more. Expect the first real run to surface something.

**Manually confirmed live,** outside the suite: I ran `dt-evals validate` against the fixture tenant — connection, spans-bucket read, and the Bedrock judge were all reachable. A `run --dry-run` correctly fetched all 192 real fixture spans. One real `run` made an actual judge call and wrote a bizevent back, which I then confirmed in Grail via DQL. So the pipeline the suite exercises works end-to-end — the suite's own assertions just haven't run yet.


test/e2e/.env.sample
```diff
@@ -0,0 +1,103 @@
+# dt-evals E2E suite — local configuration
```
yassine-dt: Probbaly delete this since this test is for us only not for users and has too much text

test/e2e/README.md
```diff
@@ -0,0 +1,263 @@
+# dt-evals E2E Tests
```
yassine-dt, line 1: Same comment here I see the E2E tests as an automatic quality gate will never be used by any user so we need to keep it lean

.gitignore
```diff
@@ -3,5 +3,27 @@ dist/
...
 
+# Local eval config written by `dt-evals configure` in the working directory.
+# It carries a live tenant API token, so it must never be committed. Matches the
+# exact filename at any depth — the tracked examples
+# (dt-eval-cli/example.dt-eval.yaml, evals/*.dt-eval.yaml) are deliberately unaffected.
+.dt-eval.yaml
+
+# Local planning artifacts (agent-skills plan/build workflow). Anchored to the
+# repo root so a legitimate source directory named `tasks/` is not swept up.
+/tasks/
```
yassine-dt: slope

.gitignore
```diff
@@ -3,5 +3,27 @@ dist/
 *.tsbuildinfo
 .env
 .env.*
+# ...but the committed templates that document which variables to set are not secrets.
+!.env.sample
```
yassine-dt, line 7: we don't need this

.github/workflows/e2e-pr-checks.yml
```diff
@@ -0,0 +1,76 @@
+---
```
yassine-dt, line 1: can we rename the files like now they are confusing. Maybe `e2e-pr-checks.yml ` and `live_weekly_e2e.yml`

test/e2e/.env.sample
armin-sams-dt: will trim down a bit to match convention from instrumentation repo from anna and flo

test/e2e/README.md
armin-sams-dt: will trim down a bit to match convention from instrumentation repo from anna and flo

.github/workflows/e2e-pr-checks.yml
armin-sams-dt: already commited. soon to be pushed

.gitignore
armin-sams-dt: as seen in the other comment. i'd rather stick to the conventions oof the instrumentation repo

.gitignore
armin-sams-dt: done

.github/workflows/e2e-pr-checks.yml
```diff
@@ -0,0 +1,97 @@
...
+# The suite is invoked whole rather than as a list of files, and the skipping is
+# what makes that safe. Naming the two credential-free files would freeze this
+# job at today's coverage: a suite added later would run nowhere, silently.
+# Running everything means a new credential-free test is gated automatically,
+# and a new credentialed one skips on its own.
+#
+# E2E_REQUIRE_ENV is deliberately NOT set here. It is what turns a missing
+# credential into a red run, and it belongs in live_weekly_e2e.yml where credentials are
+# actually expected — here, skipping is the whole point.
+on:
```
thisthat, line 7: should we also run these on a `main` commit? 🤔 


```suggestion
on:
  push:
    branches:
      - 'main'
```

armin-sams-dt: can we directly push to main? because if not, and we always go the PR way, then your suggerstion would be run twice on a merge.

yassine-dt (dismissed):
LGTM

.github/workflows/e2e-pr-checks.yml
armin-sams-dt: ok.

Let's take the suggestion. It does technically run twice on merge, but that's not pure redundancy: 
- pull_request checks out the synthetic merge commit 
- (refs/pull/:n/merge), not what actually lands on main — for squash/rebase merges that's a different commit than what push would test. 
- Job costs nothing (credential-free, 15min), so the double run is cheap insurance either way.

test/e2e/suites/validate.e2e.test.ts
```diff
@@ -0,0 +1,199 @@
...
+  destination: '✓ destination connection',
+  provider: 'Evaluator provider reachable',
+  evaluators: 'built-in evaluators available',
+} as const;
+
+/** The matching failure line for each probe, kept as its own map rather than derived — the wording varies. */
+const PROBE_FAILURE: Record<keyof typeof PROBE_SUCCESS, string> = {
+  schema: 'Config schema invalid',
+  origin: 'origin connection failed',
+  spansBucket: 'origin cannot read spans bucket',
```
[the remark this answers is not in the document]
armin-sams-dt: Confirmed gap, but a deliberate one — not something this PR should reopen.

We removed the dedicated failure case for this probe in 0068802 (`test(e2e): drop the spans-bucket misscoped-token case`). Provoking it needs a *second*, manually-provisioned tenant token with `storage:spans:read` deliberately missing — the only credential in the suite that can't route through `E2E_REQUIRE_ENV`, and the only gate whose negative case needs its own dedicated secret. We decided that wasn't worth carrying for one probe.

The probe's success path is still covered here, and the Grail SUCCEEDED-with-zero-records quirk it exists to catch is documented in the README's Troubleshooting table (`origin cannot read spans bucket`). The design doc's "What we check" section now calls out this exception explicitly, next to the existing `vertex` bad-path gap.

Happy to revisit if we decide the dedicated token is worth provisioning after all.

yassine-dt (approved):
LGTM

thisthat (approved):

PR #178, opened 2026-08-07
feat(dt-ai-ingest)!: lean async BizEvents SDK with file ingest and Parquet support

yassine-dt:
## Summary

- **Rewrite** dt-ai-ingest as a lean, async, send-only client — one `gen_ai.evaluation.result` BizEvent per eval, zero OTel tracing, zero framework coupling
- **Align schema** with dt-eval-cli's BizEvent format (`gen_ai.eval.client`, `gen_ai.response.id`, `timestamp`, `model_provider`, `service_name`, correct scoring ranges)
- **File ingest** from `.csv`, `.jsonl`, `.json`, and `.parquet` via `ingest_file()` — column names matching `Eval` fields need no `mapping=`
- **`dataset_id`** (`dt.eval.dataset_id`) — auto-generated UUID per `ingest_file()` call so every batch is queryable with `group by dt.eval.dataset_id` in DQL; caller can pass a stable label instead
- **Parquet support** via optional `pyarrow` extra (`pip install dt-ai-ingest[parquet]`) — row groups streamed, native types preserved, nulls dropped cleanly
- **Lint / markdownlint** — ruff E501 fixed, MD040 fixed, `examples/` excluded from ruff scope, `pyarrow` added to mypy ignore list

## Test plan

- [ ] `uv run pytest` — 25/25 pass (includes Parquet round-trip and null tests)
- [ ] `uv run ruff check .` — no errors
- [ ] `uv run mypy src` — no errors
- [ ] `make markdownlint` — no errors on committed MD files
- [ ] Live ingest verified against `mho70695.apps.dynatrace.com` for CSV, JSONL, and Parquet datasets


yassine-dt:
We support cdv, json, jsonl and optional parquet 
image

yassine-dt:
I added `dataset_id` so we can identify each row in his dataset context

dt-ai-ingest/src/dt_ai_ingest/_transport.py
```diff
@@ -0,0 +1,146 @@
...
+        timeout: float = 30.0,
+        chunk_size: int = 100,
+        max_retries: int = 3,
+        http_client: httpx.AsyncClient | None = None,
+    ) -> None:
+        self.endpoint = endpoint.rstrip("/")
+        self.token = token
+        self.timeout = timeout
+        self.chunk_size = chunk_size
+        self.max_retries = max_retries
```
armin-sams-dt, line 53: If max_retries is given as 0 when instantiating the Dynatrace Client, we get range(1,1) and no try will run.
Line 100: `for attempt in range(1, self.max_retries + 1):`.
This would then always throw an AssertionError `assert last_exc is not None`.
I.E if max_retries is set to 0, there won't be even a single request, despite retries suggesting, there won't only be no retry.

Suggested simnple fix:

```suggestion
        if max_retries < 1:
            raise ValueError("max_retries must be >= 1")
        self.max_retries = max_retries
```

dt-ai-ingest/src/dt_ai_ingest/_transport.py
```diff
@@ -0,0 +1,146 @@
...
+        if len(chunks) == 1:
+            await self._send_chunk(chunks[0])
+            return
+        sem = asyncio.Semaphore(_MAX_CONCURRENT_CHUNKS)
+
+        async def _guarded(chunk: list[dict[str, Any]]) -> None:
+            async with sem:
+                await self._send_chunk(chunk)
+
+        await asyncio.gather(*[_guarded(chunk) for chunk in chunks])
```
armin-sams-dt: Ok, so if one of the chunks throws an error, gather() immediately throws an exception but the other chunks continue to run. You wouldn't know how many chunks are ingested, and after a retry even be double ingested.

`return_exceptions=True` will let all chunks finish before an error is thrown.
Also, added log messages to see if chunks failed.



```suggestion
        results = await asyncio.gather(*[_guarded(chunk) for chunk in chunks], return_exceptions=True)
        errors = [e for e in results if isinstance(e, BaseException)]
        if errors:
        logger.error("%d/%d chunk(s) failed", len(errors), len(chunks))
        raise errors[0]
```

armin-sams-dt (changes requested):
I added some commits. Please review them @yassine-dt 
Furthermore you'll find 2 comments / suggestions in `_transport.py`.

armin-sams-dt (approved):
Approved after @yassine-dt addet the proposed changes.

PR #182, opened 2026-08-11
fix: follow-up bugfixes for the E2E CI run

armin-sams-dt:
## Summary
Follow-up to #177, collecting post-merge E2E CI bugfixes.

- **Timeout scaling**: `run --ci` tests had timeouts hardcoded for the 24h default lookback. Since `b2d5705` made sampling uncapped, a `workflow_dispatch` with a wider `E2E_FIXTURE_LOOKBACK` (up to the 7d cap) evaluates proportionally more spans and reliably blew past the 180s spawn timeout, even though it stayed under `dql.ts`'s 1000-span cap. `runCiTimeoutMs()` now budgets the timeout from the current lookback, judge concurrency, and seeding rate instead of a fixed number.
- **OAuth instead of a pasted token**: the suite read a pre-issued `DT_API_TOKEN` directly from the environment. It now fetches its own token via an OAuth2 client-credentials grant (`E2E_DT_OAUTH_CLIENT_ID`/`E2E_DT_OAUTH_SECRET`), scoped to what the suite's own DQL reads and `run --ci`'s writes need, and hands that token to both its own DynatraceClient and the CLI under test (unchanged, still via `DT_API_TOKEN`).
- **Sprint SSO host**: the tenant is sprint/labs tier, which has its own SSO (`sso-sprint.dynatracelabs.com`) separate from production's `sso.dynatrace.com`. An OAuth client created in a sprint tenant doesn't exist in production SSO's client database — posting there fails as a generic "ClientId or clientSecret is not correct" rather than naming the real cause. `E2E_DT_OAUTH_URL` now defaults to the sprint host in CI (a repo variable, not a secret — it's just a hostname).

### Follow-up: code review fixes
A review of the above turned up five gaps the OAuth/timeout switch introduced, all fixed here:

- **Secret-leak detection had gone blind to the real token**: `assertNoSecrets` looked for `DT_API_TOKEN` in this process's own environment, but the OAuth-fetched token is only ever handed to the CLI child's environment — it's never set here. The check was silently a no-op for the one credential that matters most. It now checks the actual value handed to the child.
- **`evallogic.e2e.test.ts` wasn't scaled**: only `run.e2e.test.ts` got the new `runCiTimeoutMs()` budget, but the `verdict` job shares the same widenable `E2E_FIXTURE_LOOKBACK` input. A wide dispatch could still time out the verdict lane.
- **Timeout budget could crash a credential-less local run**: `runCiTimeoutMs()` was called eagerly as an `it(...)` timeout argument, which vitest evaluates during test collection even for a skipped suite. A malformed `E2E_FIXTURE_LOOKBACK` value could crash the whole file for a contributor with no tenant credentials, who should just get a clean skip. It's now only computed when the suite is actually enabled.
- **Per-judge-call budget was too tight**: `JUDGE_CALL_CEILING_MS` assumed 10s per call, but the CLI's own defaults allow up to 30s x 3 attempts (with retries) for a single call — a legitimate retry could blow the computed budget. Sized off the CLI's real defaults instead.
- **OAuth token fetch had no timeout and never refreshed**: unlike the rest of the suite's `fetchWithTimeout` pattern, the token request could hang indefinitely, and the fetched token was cached for the whole process lifetime with no expiry check — a real risk now that some suites run for several minutes. Both fixed.

## Test plan
- [x] `npx tsc --noEmit` in `test/e2e` passes
- [x] Verified locally against the real tenant: `tenant`, `cli`, `contract`, `validate` (real tenant + real Bedrock judge), and the full `run` suite including the write-back round trip — all passing
- [x] Dispatched `E2E — dt-evals CLI` on this branch with the configured secrets/variable — all 64 tests plus the verdict lane passed in CI ([run 31508497520](https://github.com/dynatrace-oss/dt-evals/actions/runs/31508497520))
- [x] Re-ran locally against the real tenant after the follow-up fixes — all 65 tests passing, including the newly-scaled `evallogic.e2e.test.ts`
- [x] Re-dispatched `E2E — dt-evals CLI` on this branch [run 31511382999](https://github.com/dynatrace-oss/dt-evals/actions/runs/31511382999) after the follow-up fixes — `e2e` job green in 3m23s, `Verdict suite` green in 1m8s

### DIFF
--- a/.github/workflows/live_weekly_e2e.yml
+++ b/.github/workflows/live_weekly_e2e.yml
@@ -82,6 +82,10 @@ jobs:
           # OAuth client-credentials, not a pasted platform token — the suite fetches its own access token (env.ts).
           E2E_DT_OAUTH_CLIENT_ID: ${{ secrets.E2E_DT_OAUTH_CLIENT_ID }}
           E2E_DT_OAUTH_SECRET: ${{ secrets.E2E_DT_OAUTH_SECRET }}
+          # The tenant above is sprint/labs, not production — its OAuth client only exists on
+          # that tier's own SSO, not sso.dynatrace.com (env.ts's default). Override via repo
+          # variable, not a secret: this is a hostname, not a credential.
+          E2E_DT_OAUTH_URL: ${{ vars.E2E_DT_OAUTH_URL || 'https://sso-sprint.dynatracelabs.com/sso/oauth2/token' }}
           AWS_ACCESS_KEY_ID: ${{ secrets.E2E_AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.E2E_AWS_SECRET_ACCESS_KEY }}
           SUITE: ${{ inputs.suite }}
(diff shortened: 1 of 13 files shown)


thisthat (approved):
LGTM 👍

PR #183, opened 2026-08-12
feat(dt-ai-ingest): add integrations layer base protocol (AI-406)

yassine-dt:
## What

Ships the integration layer foundation for `dt-ai-ingest` — a typed protocol, field routing utilities, and the schema changes that make the library production-ready as a first-class eval ingestion SDK.

### Schema changes

- **`dataset_id` removed** — Grail is immutable; there is no mutable dataset registry, so the field was meaningless. Grouping is now done exclusively via `dt.eval.run_id` in DQL.
- **`score` or `label` required** — `Eval(name="x")` now raises `ValueError`. An eval with neither a numeric score nor a categorical label is not a valid observation.
- **`run_id` auto-generated** — if the caller doesn't supply one, a UUID is stamped at `ingest_file()`, `ingest()` (per-call fallback), and `EvaluationScope.__aenter__()`. Every BizEvent in Grail is always groupable by `dt.eval.run_id`.

### Integration layer (`integrations/`)

Adds `src/dt_ai_ingest/integrations/` with the base protocol all framework adapters must satisfy:

| Symbol | Purpose |
| ------ | ------- |
| `ConvertAdapter` | `@runtime_checkable` Protocol — `(result, *, run_id, mapping, defaults) → list[Eval]` |
| `require(package, extra)` | Raises helpful `ImportError` with install hint if optional dep is missing |
| `nan_to_none(v)` | Converts pandas `NaN` → `None` so Pydantic sees an absent value |
| `build_eval_kwargs(row, mapping, defaults)` | Renames columns via `mapping=`, splits known vs unknown fields, routes unknowns to `extra` |

### `pyproject.toml` extras

```toml
[project.optional-dependencies]
parquet  = ["pyarrow>=14"]
mlflow   = ["mlflow>=2.14"]
ragas    = ["ragas>=0.2"]
deepeval = ["deepeval>=0.20"]
all      = ["pyarrow>=14", "mlflow>=2.14", "ragas>=0.2", "deepeval>=0.20"]
```

Mypy overrides added for all optional framework packages so the type checker doesn't error on missing stubs.


dt-ai-ingest/src/dt_ai_ingest/integrations/_base.py
```diff
@@ -0,0 +1,52 @@
...
+def require(package: str, extra: str | None = None) -> None:
+    try:
+        __import__(package)
+    except ImportError:
+        raise ImportError(
+            f"`{package}` is required. Install with: pip install dt-ai-ingest[{extra or package}]"
+        ) from None
+
+
+def nan_to_none(v: Any) -> Any:
```
yassine-dt, line 33: To handle pandas Nan

armin-sams-dt (changes requested):
Nice work getting the base protocol in. Two things I'd want sorted before this merges, though:

**1. The `scoring_format` rename breaks parity between the two clients**

`schema.py` renames `score_1_to_5` to `rubric`. But `dt-eval-cli/src/dt/bizevent.ts:23` still emits `score_1_to_5` for the same 1-5 scale, and it wasn't updated here. The comment you removed from `schema.py` said it plainly: "Only the two formats dt-eval-cli emits are accepted, so both clients render identically."

The practical effect: any DQL query or dashboard filtering on `gen_ai.evaluation.scoring_format` will now only catch half the relevant evals, depending on which client wrote them. And nothing errors — it just quietly splits the data.

Can we either update `bizevent.ts` in this same PR, or keep `score_1_to_5` as an accepted alias in `SCORING_RANGES` until both clients agree on the new name?

**2. The README promises things that aren't actually in this diff**

The PR description and README talk about new `pyproject.toml` extras (`parquet`, `mlflow`, `ragas`, `deepeval`, `all`) with mypy overrides, and an `integrations/mlflow.py` module with a `from_mlflow` example. None of that is in the diff — only `integrations/_base.py` and an empty `__init__.py` landed, and `pyproject.toml` is untouched.

Anyone who follows the README as written will hit a `pip install` error or `ModuleNotFoundError`.

Could we trim the description/README down to what's actually shipping here (just the base protocol), and save the extras and adapters for a follow-up PR? Or, if they were meant to land together, let's add them now.



There are also 3 inline suggestions on this PR worth checking.

dt-ai-ingest/src/dt_ai_ingest/integrations/_base.py
```diff
@@ -0,0 +1,52 @@
...
+    try:
+        return None if math.isnan(v) else v
+    except TypeError:
+        return v
```
armin-sams-dt, line 37: `math.isnan(v)` blows up on pandas' `pd.NA` (and `NaT`) with a `TypeError`, so those slip through untouched instead of turning into `None`. Using `v != v` instead catches the whole NaN family, `pd.NA` included, without pulling in pandas.

```suggestion
def nan_to_none(v: Any) -> Any:
    try:
        if v != v:  # NaN-family sentinels (float NaN, pandas NA, NaT) are never equal to themselves
            return None
        return v
    except TypeError:
        return v
```

dt-ai-ingest/src/dt_ai_ingest/integrations/_base.py
```diff
@@ -0,0 +1,52 @@
...
+    ) -> list[Eval]: ...
+
+
+def require(package: str, extra: str | None = None) -> None:
+    try:
+        __import__(package)
+    except ImportError:
+        raise ImportError(
+            f"`{package}` is required. Install with: pip install dt-ai-ingest[{extra or package}]"
+        ) from None
```
armin-sams-dt, line 30: The `extra or package` fallback assumes the pip extra name matches the import name, but that's wrong for `pyarrow` — the actual extra is `parquet`. Calling `require("pyarrow")` without passing `extra="parquet"` would print an install hint for an extra that doesn't exist, so let's just require callers to pass it.

```suggestion
def require(package: str, extra: str | None = None) -> None:
    try:
        __import__(package)
    except ImportError:
        raise ImportError(
            f"`{package}` is required. Install with: pip install dt-ai-ingest[{extra}]"
        ) from None
```

dt-ai-ingest/README.md
```diff
@@ -121,39 +126,42 @@ async with dt_ai_ingest.evaluation(run_id="run-1") as record:
...
+Unknown kwargs passed directly to `Eval()` are rejected — use `extra={"custom.key": "value"}` to attach arbitrary keys. When ingesting from files or framework integrations, unrecognised columns are automatically routed to `extra`.
 
 | Field | BizEvent key | Meaning |
 | ----- | ------------ | ------- |
 | `name` | `gen_ai.evaluation.name` | Metric name, e.g. `faithfulness`. **Required.** |
-| `score` | `gen_ai.evaluation.score.value` | Numeric score. |
-| `label` | `gen_ai.evaluation.score.label` | Categorical outcome, e.g. `pass` / `fail`. |
-| `scoring_format` | `gen_ai.evaluation.scoring_format` | `score_0_to_1` (default) or `score_1_to_5`. |
+| `score` | `gen_ai.evaluation.score.value` | Numeric score. **Required.** |
+| `label` | `gen_ai.evaluation.score.label` | Categorical outcome, e.g. `pass` / `fail`. Optional. |
```
armin-sams-dt, line 138: `label` is now locked to `Literal["pass", "fail"]` and raises a `ValueError` otherwise, but the docs still present it as a loose example. Anyone using another categorical value that worked before this PR will hit a validation error with no warning here.

```suggestion
| `label` | `gen_ai.evaluation.score.label` | Categorical outcome. Must be `pass` or `fail`. Optional. |
```

yassine-dt: Good point I will go through teh Readme

dt-ai-ingest/src/dt_ai_ingest/integrations/_base.py
yassine-dt: This is for ragas etc no pyarraow

dt-ai-ingest/src/dt_ai_ingest/integrations/_base.py
yassine-dt: this won't catch pd.NA either

armin-sams-dt (approved):

PR #186, opened 2026-08-12
fix: bump vitest 2.x -> 4.x to clear devDependency vulnerabilities

armin-sams-dt:
## Summary
- `npm audit` in `test/e2e`, `dt-eval-cli`, and `dt-eval-deploy` flags a critical (GHSA-5xrq-8626-4rwp in vitest) plus several high/moderate findings in the vitest/vite/esbuild test toolchain. `dt-eval-lib` was already on `vitest ^4.1.4`; this brings the other three in line.
- All affected packages are devDependencies of the test toolchain — never shipped in production code — and the critical only applies when the Vitest UI server (`--ui`) is running, which we never do in CI or locally.
- Vitest 4 removed `poolOptions.forks.singleFork`; `test/e2e/vitest.config.ts` already sets `fileParallelism: false`, which the new docs say pins `maxWorkers` to 1 — same effect, so the now-invalid option is just dropped.
- As a side effect of refreshing the lockfiles, `dt-eval-cli`/`dt-eval-deploy` also lost their critical `protobufjs` finding (npm re-resolved to a patched version already inside `@google/genai`'s existing range). Remaining high/low findings there are a separate, pre-existing production dependency issue (protobufjs/ws/form-data via `@google/genai`/`@anthropic-ai/sdk`) and are intentionally out of scope for this PR.

## Test plan
- [x] `npm run typecheck` in `test/e2e` passes
- [x] `npx vitest run` in `test/e2e` — all 65 credential-free tests pass
- [x] `npm run test` in `dt-eval-cli` — same single pre-existing failure (`probeProvider — bedrock ...`) reproduced on `main` before this change, confirmed not a regression
- [x] `npm run build` in `dt-eval-lib` and `dt-eval-cli` succeed
- [x] `npm audit` — 0 vulnerabilities in `test/e2e`; vitest/vite/esbuild findings cleared in `dt-eval-cli`/`dt-eval-deploy`


armin-sams-dt (commented):
Only bumping vitest to a higher version.

yassine-dt (dismissed):

thisthat (approved):

PR #190, opened 2026-08-12
feat(dt-ai-ingest): add Ragas EvaluationResult adapter (AI-411)

yassine-dt:
## What

Adds `from_ragas`, a convert-style adapter that turns a Ragas `EvaluationResult`
into `Eval` rows — one `gen_ai.evaluation.result` BizEvent per `(sample, metric)`
pair, carrying the sample's question/answer.

> **Stacked on #183** (base branch `fix/AI-406-refactor-adapters-for-ai-ingest`,
> the integrations base layer). Please review/merge #183 first — the diff here is
> just the ragas commit on top.

## Design

- **Never imports `ragas` or `langchain`.** The adapter duck-types on the public
  `result.scores` / `result.to_pandas()` surface, so importing the module is cheap
  and `mypy --strict` passes without the extra installed.
- Satisfies the `ConvertAdapter` contract from the base layer (AI-406).
- Maps `user_input → question` and `response → answer` by default; large/noisy
  columns like `retrieved_contexts` are dropped unless the caller maps them in.
- `run_id` and `defaults` (e.g. judge `model`/`model_provider`) are stamped on
  every produced Eval.

## Included

- `ragas` optional extra (`pip install dt-ai-ingest[ragas]`) + locked deps.
- Dependency-free unit tests (a fake `EvaluationResult`) so CI runs them under
  plain `uv sync` — no heavy ragas install required.
- Auto-discovered **conformance test**: asserts every `from_*` adapter under
  `integrations/` matches the `ConvertAdapter` shape. This is the CI backstop now
  that `@runtime_checkable` is gone (a Protocol only checks "is callable" at
  runtime, not the signature).
- README: real Ragas integration section.

## Verified end-to-end

Ran a **real** Ragas 0.2.15 evaluation (metrics: `faithfulness` +
`llm_context_precision_without_reference`, both LLM-only) over a 3-sample RAG
dataset and ingested the results into a live Dynatrace tenant:

- `result.scores` really is `list[dict[metric → score]]`, and `to_pandas()`
  concatenates dataset input columns + one column per metric — matching the
  adapter's assumptions.
- 3 samples × 2 metrics fanned out into 6 Evals; a deliberately context-
  contradicting answer correctly scored `faithfulness = 0.0`.
- 6 BizEvents ingested and groupable in Grail by `dt.eval.run_id`.

## CI

`ruff check .` clean · `mypy --strict src` clean (10 files) · `pytest` 49 passed /
2 skipped — verified in a pristine checkout (committed files only).


dt-ai-ingest/src/dt_ai_ingest/integrations/ragas.py
```diff
@@ -0,0 +1,83 @@
...
+    *,
+    run_id: str | None = None,
+    mapping: dict[str, str] | None = None,
+    defaults: dict[str, Any] | None = None,
+) -> list[Eval]:
+    """One Eval per (sample, metric); NaN scores skipped.
+
+    Binary metrics (Ragas ``AspectCritic``) also carry a ``pass``/``fail`` label;
+    metrics scored above 1 use the ``rubric`` scoring_format.
+    """
```
armin-sams-dt, line 56: The explicit `run_id` parameter loses to `defaults["run_id"]` here. If a caller passes both, `setdefault` keeps whatever's already in `defaults` and ignores the parameter — backwards from what you'd expect, since the explicit argument should win.

```suggestion
    if run_id is not None:
        base_defaults["run_id"] = run_id
```

armin-sams-dt (dismissed):

armin-sams-dt (approved):

PR #196, opened 2026-08-14
feat(dt-eval-lib): add deterministic evaluators (exact_match, regex, must_not_match, json_schema, must_contain, must_not_contain)

yassine-dt:
## What

Adds **deterministic (code-based) evaluators** to `dt-eval-lib`, alongside LLM-as-a-judge. A `method` field on `PromptDefinition` picks the scorer; all return the same `EvalResult`, so nothing downstream changes. No `method` ⇒ `llm_as_judge` (backward compatible), and deterministic methods need no provider.

| method | passes when |
|---|---|
| `exact_match` | output equals the expected value |
| `regex` | pattern matches the output |
| `must_not_match` | pattern does **not** match |
| `json_schema` | output parses as JSON and validates |
| `must_contain` | keyword(s) present (`mode: any\|all`) |
| `must_not_contain` | keyword(s) absent |

## Limitations / out of scope

- `json_schema` needs `ajv`, and `regex`/`must_not_match` need `recheck`; both are **optional peers** (lazy-loaded) so an LLM-only install stays lightweight — those methods error if the peer isn't installed.
- Regex input is bounded to 100k chars, and patterns are rejected unless proven free of catastrophic backtracking (ReDoS).


dt-eval-lib/src/engine/deterministic/contains.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+export function contains(
+  input: EvalInput,
+  params: ContainsParams,
+  mustContain: boolean,
+): DeterministicOutcome {
+  const keywords = params.keywords ?? [];
+  if (keywords.length === 0) {
+    throw new EvalConfigError(
+      "must_contain/must_not_contain requires a non-empty 'keywords' array",
```
thisthat: NIT:
```suggestion
      "${mustContain ? 'must_contain' : 'must_not_contain'} requires a non-empty 'keywords' array",
```

dt-eval-lib/src/engine/deterministic/contains.ts
```diff
@@ -0,0 +1,46 @@
+import { EvalConfigError } from "../../errors";
+import type { ContainsParams } from "../../prompts/types";
```
thisthat: mmm I dunno if these types should live here. I would see it more on the `types.ts` file in this folder 🤔

yassine-dt: yes good catch stupid Claude  and stupid me missed this ahahah

dt-eval-lib/src/engine/evaluate.ts
```diff
@@ -2,25 +2,37 @@ import { EvalConfigError, EvalInputError, EvalResponseError, EvalTimeoutError }
...
 import { createProvider } from "./providers/index";
 import type { LLMJudgeResponse } from "./providers/types";
 import type { EvalConfig, EvalInput, EvalResult } from "./types";
 
 /**
  * Main evaluation function.
- * Resolves the metric, validates input, renders the prompt,
- * calls the LLM provider, and computes the final score.
+ * Resolves the metric and validates input, then either runs a deterministic
+ * scorer or renders the prompt and calls the LLM provider.
```
armin-sams-dt: Keep this description method-agnostic so it stays accurate when new evaluation methods are added.
```suggestion
 * Resolves the metric, validates input, and dispatches to the configured
 * evaluation method.
```

dt-eval-lib/src/engine/types.ts
```diff
@@ -37,7 +37,8 @@ export interface ScoringOptions {
 
 /** Top-level evaluation configuration */
 export interface EvalConfig {
-  provider: ProviderOptions;
+  /** Required for LLM-judge evaluators; unused by deterministic methods */
```
armin-sams-dt: This comment can be removed.
```suggestion
```

dt-eval-lib/src/prompts/types.ts
```diff
@@ -20,15 +21,29 @@ export enum BuiltInMetric {
 /** Identifier for the population-level drift detection metric. */
 export const DRIFT_METRIC_ID = "drift";
 
+/** How an evaluator computes its score. Defaults to "llm_as_judge" when omitted. */
```
armin-sams-dt: Clarify that the default is applied by the evaluation engine rather than by the type itself.
```suggestion
/** Evaluation method. The evaluation engine defaults to "llm_as_judge" when omitted. */
```

dt-eval-lib/src/prompts/types.ts
```diff
@@ -20,15 +21,29 @@ export enum BuiltInMetric {
...
 export interface PromptDefinition {
   id: string;
   name: string;
   version: string;
   /** Description of what this metric evaluates */
   description: string;
-  /** The evaluation prompt template — uses {{input}}, {{output}}, {{context}}, {{expectedOutput}} placeholders */
-  prompt: string;
-  /** Which input fields this prompt requires */
+  /** Evaluation method — defaults to "llm_as_judge" when omitted */
```
armin-sams-dt: Same clarification here for consistency.
```suggestion
  /** Evaluation method. The evaluation engine defaults to "llm_as_judge" when omitted. */
```

armin-sams-dt (approved):
Approved:
* review by hand for the most critical parts
* review by agentic code review and adversary models

PR #197, opened 2026-08-14
feat(dt-eval-cli): wire deterministic evaluators (config, runner, telemetry)

yassine-dt:
## What

Lets the deterministic evaluators from #196 be **authored in `.dt-eval.yaml` and run** in the normal pipeline, with the evaluation surfaced in telemetry. A metric entry gains optional `method` + `params` (omit `method` for LLM-judge); `MetricInputs` gains `expectedOutput` routing; `schemaVersion` → 3. Config is validated at load time, deterministic input is built the same way as for LLM-judge, and bizevents record the real `gen_ai.evaluation.method` — with `gen_ai.evaluation.type` = `custom` for user-authored deterministic metrics vs `ready_made` for built-in LLM-judge, and judge `provider`/`model` written only for LLM metrics.

```yaml
metrics:
  enabled:
    - toxicity                        # llm_as_judge
    - id: valid-json
      method: json_schema
      params: { schema: { type: object, required: [answer] } }
    - id: no-refusal
      method: must_not_contain
      params: { keywords: ["i cannot help", "i'm sorry"], mode: any }
```

## Limitations / out of scope

- **Release ordering:** the CLI consumes `dt-eval-lib` from the registry — bump `@dynatrace-oss/dt-eval-lib` (and the lockfile) after #196 is released.
- Span content is PII-masked before evaluation, so a "no PII" guardrail detects the mask marker (`must_not_contain: ["[REDACTED]"]`) rather than raw values.
- A deterministic entry has only an `id` (no separate display name), so `gen_ai.evaluation.name` equals `gen_ai.evaluation.spec_id`.
- A user-defined *custom LLM-judge* prompt is still tagged `ready_made` (type is derived from method); only deterministic methods are marked `custom`.
- Authoring is YAML-only (no interactive wizard yet).
- `dt-eval-deploy` needs `recheck` added to run regex methods in serverless; without it those methods error rather than run.

### DIFF
--- a/dt-eval-cli/package.json
+++ b/dt-eval-cli/package.json
@@ -23,6 +23,7 @@
   "scripts": {
     "build": "tsup",
     "test": "vitest run",
+    "typecheck": "tsc --noEmit",
     "dev": "tsx src/index.ts"
   },
   "homepage": "https://github.com/dynatrace-oss/dt-evals#readme",
--- a/dt-eval-cli/package.json
+++ b/dt-eval-cli/package.json
@@ -41,10 +41,12 @@
     "@google/genai": "^1.0.0",
     "@inquirer/prompts": "^7.10.1",
     "@types/figlet": "^1.7.0",
+    "ajv": "^8.20.0",
     "commander": "^12.0.0",
     "figlet": "^1.11.0",
     "openai": "^4.73.0",
     "picocolors": "^1.1.1",
+    "recheck": "^4.5.0",
     "yaml": "^2.4.0"
   },
   "devDependencies": {
--- a/.github/workflows/e2e-pr.yml
+++ b/.github/workflows/e2e-pr.yml
@@ -56,6 +56,7 @@ jobs:
         run: |
           npm ci
           npm install ../dt-eval-lib
+          npm run typecheck
           npm run build
 
       - name: Install E2E dependencies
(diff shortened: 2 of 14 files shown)


dt-eval-cli/package.json
```diff
@@ -41,10 +41,12 @@
     "@google/genai": "^1.0.0",
     "@inquirer/prompts": "^7.10.1",
     "@types/figlet": "^1.7.0",
+    "ajv": "^8.20.0",
```
thisthat, line 45: where is this used? 🤔 I couldn't find it any ref to this

yassine-dt: we use it to check json schema 
https://github.com/dynatrace-oss/dt-evals/blob/feat/deterministic-evals-lib/dt-eval-lib/src/engine/deterministic/json-schema.ts#L21 
Let me know WDYT Idk if they are better ways to do it in TS

armin-sams-dt: `ajv` is a runtime peer dependency of `@dynatrace-oss/dt-eval-lib` used by the `json_schema` evaluator. Since the library declares it optional, the CLI must provide it for YAML-authored JSON Schema metrics. The PR now consumes the released `dt-eval-lib@0.0.16-alpha`.

armin-sams-dt (dismissed):

armin-sams-dt (approved):
only rebase

PR #207, opened 2026-08-26
docs(dt-eval-cli): add deterministic evaluator examples (AI-454)

yassine-dt:
## What

Documents the new deterministic (code-check) evaluators in the CLI README, under a new **Deterministic evaluators (code checks)** subsection in `## Built-in Evaluators`.

General, provider-agnostic examples (no project-specific data):

- `must_contain` — require an expected keyword
- `must_not_contain` — block forbidden/harmful words (placeholder tokens)
- `must_not_match` — fail if the output matches a pattern (e.g. a leaked IP address), with a pointer to `regex` for the inverse
- `json_schema` — require valid, well-structured JSON

Includes a note that `json_schema` needs **bare** JSON (markdown-fenced JSON won't parse) and that `json_schema` / `regex` / `must_not_match` rely on the optional `ajv` / `recheck` dependencies.

## Why

The deterministic evaluators shipped without README coverage. Examples were validated end-to-end against a live tenant before writing them.

## Notes

- Docs-only change; `markdownlint` passes clean.

ref: AI-454

### DIFF
--- a/dt-eval-cli/README.md
+++ b/dt-eval-cli/README.md
@@ -606,6 +606,68 @@ Drift results are written back as the same event type with `gen_ai.evaluation.ty
 | `toxicity` | Harmful, offensive, or unsafe output | `output` |
 | `user-frustration` | Frustration signals in the user's message | `input` |
 
+### Deterministic evaluators (code checks)
+
+Not every check needs an LLM. Add a `method` to a metric entry in
+`metrics.enabled` to run a fast, reproducible, zero-cost code check instead of
+an LLM judge. Each check scores the span output as pass (`1.0`) or fail
+(`0.0`). Omitting `method` keeps the default LLM-as-judge behavior, so existing
+configs are unaffected.
+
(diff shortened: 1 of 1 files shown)


dt-eval-cli/README.md
```diff
@@ -606,6 +606,68 @@ Drift results are written back as the same event type with `gen_ai.evaluation.ty
...
+      params:
+        pattern: "\\b(?:\\d{1,3}\\.){3}\\d{1,3}\\b"
+```
+
+> Use `regex` for the opposite check — pass only when the output **does** match
+> the pattern.
+
+**`json_schema` — require valid, well-structured JSON**
+
+```yaml
```
thisthat, line 657: I think for this use case we should do a follow-up with a proper in depth example 🤔

thisthat (approved):
Thanks for the docs 🙌

dt-eval-cli/README.md
yassine-dt: yes agreed I will add more in depth examples

PR #212, opened 2026-08-27
feat(dt-eval-cli): add conversation grouping and span selection for trajectory mode (AI-458-3)

yassine-dt:
## Summary
- Adds `maxConversations` to `ScopeConfig` (trajectory mode only)
- Extends `DqlQueryOptions` with `mode` and `maxConversations`; trajectory mode fetches `gen_ai.conversation.id` + `gen_ai.response.finish_reasons`, sorts by `end_time desc`, and uses a `maxConversations * 20` limit
- Adds `conversationId` and `finishReasons` to `GenAiSpan` and populates them in `parseSpanResults`
- Adds exported `selectTrajectorySpans`: groups by `conversationId` (fallback: `traceId`), picks the best span per group (prefers `stop` finish reason, then latest `endTime`/`startTime`), caps at `maxConversations`
- Runner wires `mode`/`maxConversations` into `buildGenAiSpanQuery` and calls `selectTrajectorySpans` after operation-name filtering in trajectory mode; span mode flow is unchanged

## Test plan
- [ ] Trajectory mode DQL query includes `gen_ai.conversation.id`, `gen_ai.response.finish_reasons`, sorts by `end_time desc`, and limits to `maxConversations * 20`
- [ ] Span mode DQL query is unchanged
- [ ] `selectTrajectorySpans` groups correctly by `conversationId` when present, `traceId` otherwise
- [ ] `stop` finish-reason preference and latest-time tiebreaker work as specified
- [ ] `maxConversations` cap enforced (default 200)
- [ ] Span mode runner path is unaffected

### DIFF
--- a/dt-eval-cli/src/config/schema.ts
+++ b/dt-eval-cli/src/config/schema.ts
@@ -80,6 +80,8 @@ export interface ScopeConfig {
   spanFields?: SpanFieldsMap;
   mode?: "span" | "trajectory";
   keepPartTypes?: string[];
+  /** Maximum number of conversations to evaluate. Only used in trajectory mode. */
+  maxConversations?: number;
 }
 
 /**
--- a/dt-eval-cli/src/dt/dql.ts
+++ b/dt-eval-cli/src/dt/dql.ts
@@ -12,6 +12,8 @@ export interface DqlQueryOptions {
   spanFields?: SpanFieldsMap;
   /** GenAI operation names to keep. Empty array disables this filter. */
   operationNames?: string[];
+  mode?: "span" | "trajectory";
+  maxConversations?: number;
 }
 
 export type DqlResult = GenAiSpan[];
@@ -77,9 +79,10 @@ export function filterSpansByOperationName(spans: GenAiSpan[], operationNames?: 
 }
 
 export function buildGenAiSpanQuery(opts: DqlQueryOptions): string {
-  const { app, since, limit = 1000, errorsOnly = false, spanFields } = opts;
+  const { app, since, limit = 1000, errorsOnly = false, spanFields, mode, maxConversations = 200 } = opts;
   const operationNames = resolveOperationNames(opts.operationNames);
   const fields = resolveFields(spanFields);
+  const isTrajectory = mode === 'trajectory';
 
   // Use explicit from:/to: timeframe — a filter alone leaves Grail's default
   // ~2h analysis window in effect even when the filter asks for longer.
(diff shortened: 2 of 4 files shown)


thisthat (dismissed):

armin-sams-dt:
Part of the AI-458 stack review (#215, #212, #213). This PR holds the finding I would most want addressed, though nothing here is blocking.

### Preferring `stop` over recency picks a shorter history

The JSDoc that #213 adds states the assumption: the selected span's `gen_ai.input.messages` carries the full conversation. Under that assumption the right representative is the conversation's last span. The comparator puts stop-preference above the timestamp tiebreak, so an earlier `stop` beats a later `tool_calls`.

Failure case: an eight-call agent conversation where call 5 finishes with `stop`, because the model emitted text and the orchestrator then ran three more tool rounds, and calls 6 through 8 finish with `tool_calls`. `selectTrajectorySpans` returns call 5, so the judge sees five turns of a nine-turn conversation and scores it as complete. The same inversion drops the trajectories you most want to look at, since a final turn that hits `length` or `content_filter` loses to an earlier clean `stop`.

Suggestion: make time the primary key and use `stop` only for equal timestamps, or drop the stop-preference entirely. "Latest span in the group" already says what you mean. If the goal is to exclude mid-stream spans, that is a filter, not a comparator.

### `conversationId` is parsed but never emitted

This PR adds `conversationId` to `GenAiSpan` and keys the whole feature on it, but `buildBizeventPayload` never writes it. Eval results land with `trace_id` and `span_id` only, so in Dynatrace you cannot group conversation-level results by conversation, which is the first query anyone will run against this feature. One line:

```ts
if (span.conversationId) payload['gen_ai.conversation.id'] = span.conversationId;
```

If it is deliberately deferred to a later PR in the series, worth saying so in the body.

### Nothing warns when the row limit saturates

`maxConversations * 20` degrades quietly in two ways. A conversation with more than 20 spans is fetched only in part, and the conversation straddling row 4000 loses its tail. In the second case the group's selected span is an intermediate one, so a truncated history gets judged as complete. Logging a warning when the returned record count equals the requested limit turns a silent wrong answer into a visible one.

### The grouping key mixes two ID namespaces

`span.conversationId ?? span.traceId` splits a conversation whose spans only partly carry `gen_ai.conversation.id` into one conversationId group plus one group per bare trace. That inflates the conversation count and produces partial-history "conversations". Prefixing the key with `c:` and `t:` also removes the collision between a conversation id and a trace id, however unlikely that is.

### No validation for `maxConversations`

`0` emits `| limit 0`. `-5` emits `| limit -100`, which surfaces as an opaque Grail error rather than a config error. `1e9` emits `| limit 20000000`. `validateConfig` already validates `since`, `operationNames` and `sampling.strategy`, so a positive-integer check with an upper bound fits the existing pattern.

### Smaller things

- `DqlQueryOptions.limit` is silently ignored in trajectory mode, and nothing in the type or its JSDoc says so.
- `TRAJECTORY_SPANS_PER_CONVERSATION = 20` carries real behavioural weight. One comment on where 20 came from would help the next reader.
- `runner/index.ts` logs `dropped: parsedSpans.length - allSpans.length`, which now mixes operation-name drops with conversation collapse. A normal trajectory run will report roughly 3800 spans "dropped" that were not dropped. Splitting into `dropped` and `collapsed` would keep the number honest.
- `selectTrajectorySpans` is exported for testability but has no tests. Grouping by `conversationId`, the `traceId` fallback, the stop-versus-later tiebreak, the `maxConversations` cap, and empty input would cover it.

thisthat (approved):

PR #213, opened 2026-08-27
feat(dt-eval-cli): add maxMessages truncation fallback for trajectory mode (AI-458-4)

yassine-dt:
## Summary
- Adds `maxMessages?: number` to `ScopeConfig` (trajectory mode only)
- Adds `DEFAULT_MAX_MESSAGES = 50` constant and `maxMessages?: number` to `ParseSpanOptions` in `dql.ts`
- `extractFullHistory` now accepts a `maxMessages` parameter and slices the oldest messages when the limit is exceeded
- `runner/index.ts` passes `evalConfig.scope.maxMessages` through to `parseSpanResults`

## Test plan
- [ ] Set `maxMessages: 5` in a trajectory-mode config and verify the extracted history is truncated to the last 5 messages
- [ ] Verify existing trajectory runs without `maxMessages` default to 50 messages max
- [ ] `tsc --noEmit` passes with zero errors

### DIFF
--- a/dt-eval-cli/src/config/schema.ts
+++ b/dt-eval-cli/src/config/schema.ts
@@ -80,7 +80,10 @@ export interface ScopeConfig {
   spanFields?: SpanFieldsMap;
   mode?: "span" | "trajectory";
   keepPartTypes?: string[];
-  /** Maximum number of conversations to evaluate. Only used in trajectory mode. */
+  /**
+   * Fetch/selection cap for trajectory mode: how many conversation groups to consider before
+   * `sampling` is applied (default: random @ 5%). Increase to widen coverage, decrease to cut cost.
+   */
   maxConversations?: number;
   /** Maximum number of messages to keep per trajectory (oldest are dropped). Only used in trajectory mode. */
   maxMessages?: number;
--- a/dt-eval-cli/src/config/schema.ts
+++ b/dt-eval-cli/src/config/schema.ts
@@ -82,6 +82,8 @@ export interface ScopeConfig {
   keepPartTypes?: string[];
   /** Maximum number of conversations to evaluate. Only used in trajectory mode. */
   maxConversations?: number;
+  /** Maximum number of messages to keep per trajectory (oldest are dropped). Only used in trajectory mode. */
+  maxMessages?: number;
 }
 
 /**
--- a/dt-eval-cli/src/dt/dql.ts
+++ b/dt-eval-cli/src/dt/dql.ts
@@ -24,6 +24,7 @@ const PROMPT_SLOTS = 3;
 
 const TRAJECTORY_SPANS_PER_CONVERSATION = 20;
 const DEFAULT_MAX_CONVERSATIONS = 200;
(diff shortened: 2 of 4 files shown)


thisthat (dismissed):

armin-sams-dt:
Part of the AI-458 stack review (#215, #212, #213). Nothing blocking, but the sampler change deserves a second look.

### The sampler change is out of scope, and the suite cannot catch it

`applySampling`'s `latest` strategy now sorts on `endTime ?? startTime` for every mode, so this changes span-mode behaviour inside a PR described as a trajectory-mode `maxMessages` fallback. It appears in neither the summary nor the test plan.

It is also unverifiable by the current tests, which is the part I would fix first. `makeSpan` in `tests/runner/sampler.test.ts` sets a `timestamp` field, but `applySampling` sorts on `startTime` and `endTime`, which no fixture ever sets. Both sides of the comparator evaluate to `new Date(0)`, the sort becomes a no-op, the input order survives, and the assertions pass. They would pass just as happily with the comparator inverted. `timestamp` is not even a field on `GenAiSpan`; nobody noticed because `tsconfig.json` excludes `tests` from typechecking.

So the `latest` tests have been passing for free, and this PR inherits that. A case where `startTime` and `endTime` disagree on ordering, with both actually populated, is the useful addition.

Either way, worth splitting the sampler change into its own commit or naming it in the body and the README sampling table.

### NaN path in the new comparator

`new Date(...).getTime()` returns NaN for anything that is not a parseable date. `NaN >= NaN` is false, so the reducer returns `b` unconditionally, and under `sort end_time desc` that is the oldest span in the group, therefore the shortest history. Fixtures use ISO strings so this is theoretical today, but the old lexicographic string compare had no such path. A `Number.isFinite` guard is one line.

### `maxMessages: 0` does the opposite of what it says, and nothing validates it

`maxMessages: 0` or a negative value makes `msgs.slice(...)` return `[]`, which stringifies to `'[]'`, which the guard in `parseSpanResults` then rejects, so `input` stays the full untruncated history. Same for a non-integer. I left the details on #215 since that guard is the shared cause. A positive-integer check in `validateConfig`, alongside the existing `since` and `sampling.strategy` checks, would catch it at config load.

### `maxMessages` caps count, not size

A single `tool_call_response` part can be tens of KB, a DQL result set or a large API response, so 50 messages is unbounded in tokens. The judge bill is the exposure, and with `storeEvaluatedPrompt` enabled the whole `span.input` also goes into a bizevent attribute per span per metric. A character budget applied newest-first, alongside the message cap, would make the "fallback for oversized histories" claim hold.

### Docs

`maxConversations` and `maxMessages` are user-facing config with no README entry. The JSDoc you added for `maxConversations` gets the non-obvious part right, that it is a pre-sampling cap. That belongs in the README too, because with the default `random @ 5%` a default trajectory run fetches up to 4000 spans, collapses them to at most 200 conversations, and then evaluates about 10 of them. Nobody will guess that from `maxConversations: 200`.

thisthat (approved):

PR #215, opened 2026-08-27
feat(dt-eval-cli): add trajectory mode flag and preserve full message history (AI-458-1+2)

yassine-dt:
## Summary
- Adds \`mode?: "span" | "trajectory"\` to \`ScopeConfig\` (AI-458-1, squashed in)
- Preserves full \`gen_ai.input.messages\` history in trajectory mode instead of discarding it (AI-458-2)
- Adds \`keepPartTypes\` config (default: \`["text", "tool_call", "tool_call_response"]\` — reasoning stripped by default, configurable)
- Adds \`maxMessages\` truncation fallback for oversized histories (default: 50)
- Span mode behaviour is completely unchanged

## Test plan
- [ ] Verify span mode configs produce identical output to before this change
- [ ] Verify trajectory mode passes full message history to the judge

### DIFF
--- a/dt-eval-cli/src/config/defaults.ts
+++ b/dt-eval-cli/src/config/defaults.ts
@@ -39,6 +39,7 @@ export const DEFAULT_CONFIG: Omit<DtEvalConfig, 'dynatrace' | 'judge'> & {
   },
   scope: {
     since: '1h',
+    mode: 'span' as const,
     operationNames: DEFAULT_OPERATION_NAMES,
     sampling: {
       strategy: 'random',
--- a/dt-eval-cli/src/config/index.ts
+++ b/dt-eval-cli/src/config/index.ts
@@ -390,6 +390,11 @@ export function validateConfig(config: DtEvalConfig): void {
     }
   }
 
+  const mode = config.scope?.mode as unknown;
+  if (mode !== undefined && mode !== 'span' && mode !== 'trajectory') {
+    issues.push(`scope.mode must be "span" or "trajectory" (got "${mode}")`);
+  }
+
   const { strategy } = config.scope?.sampling ?? {};
   if (strategy && !['random', 'latest', 'errors-only'].includes(strategy)) {
     issues.push(`scope.sampling.strategy must be one of: random, latest, errors-only`);
--- a/dt-eval-cli/src/config/schema.ts
+++ b/dt-eval-cli/src/config/schema.ts
@@ -79,6 +79,7 @@ export interface ScopeConfig {
   /** Custom span attribute mapping. Defaults handle OTel + OpenLLMetry. */
   spanFields?: SpanFieldsMap;
   mode?: "span" | "trajectory";
+  keepPartTypes?: string[];
 }
 
 /**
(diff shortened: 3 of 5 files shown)


thisthat (commented):
I don't see where `span` is the default mode to support backwards compatibility of config 🤔

thisthat (approved):

armin-sams-dt:
Reviewed the whole AI-458 stack (#215, #212, #213) together. The decomposition is good: three small PRs, span mode deliberately left alone. These comments are about #215 only, the other two have their own. None of it is blocking, your call on what to take.

### The `'[]'` guard hides misconfiguration

```ts
if (filtered && filtered !== '[]') input = filtered;
```

When `extractFullHistory` legitimately filters everything away, this keeps `input` as the untouched raw JSON, which is the opposite of what the config asked for.

Concrete case: an agent whose input messages hold only `tool_call` and `tool_call_response` parts, run with `keepPartTypes: ['text']`. Every message loses all its parts, the `parts.length > 0` filter drops every message, the result is `'[]'`, the guard rejects it, and the judge receives the full unfiltered history including the reasoning parts that were configured away. Same path for `keepPartTypes: []`, for `maxMessages: 0` (added in #213), and for `keepPartTypes: "text"` written as a string, where `new Set("text")` becomes `{'t','e','x'}` and matches nothing.

Two costs. Judge tokens on a payload that can be an order of magnitude larger than intended, and reasoning traces reaching the judge after someone asked to strip them.

Suggestion: separate "parse failed" from "filtered to empty". Return `null` only for unparseable input, let an empty history stay empty, and log a warning. If an empty trajectory should not be judged, skip the span rather than substituting the raw blob.

### No validation for `mode` and `keepPartTypes`

`validateConfig` already checks `scope.since` with a regex, `scope.operationNames` element by element, and `scope.sampling.strategy` against its enum, and 8043631 went further in the same direction for the deterministic evaluators. The two new fields get nothing, so `mode: trajectroy` runs span mode with no error and no log. That is the most likely typo in this feature and the least visible one.

### Tests

No tests in any of the three PRs, and every test-plan box is unchecked. `tests/dt/dql.test.ts` already covers this file with 40+ cases, so the gap is not infrastructure. Green CI here means the old tests still pass, nothing more.

For #215 specifically: default part filtering, custom `keepPartTypes`, each of the empty-result cases above, and one assertion that span mode output is unchanged.

### Docs

The README documents `operationNames`, `sampling` and `spanFields` in a YAML example plus prose. `mode` and `keepPartTypes` arrive with no README entry and no JSDoc, while every neighbouring field in `ScopeConfig` has one.

### PR description

The body lists "Adds `maxMessages` truncation fallback for oversized histories (default: 50)", but that is #213's diff. Squashed commit messages live in main forever, so it is worth trimming to what this PR actually does.

### Smaller things

- Trajectory extraction only fires when the input came from `gen_ai.input.messages`. A custom `spanFields.input` together with `mode: trajectory` silently behaves like span mode.
- `mode?: "span" | "trajectory"` is spelled out in three places, with double quotes in files that use single quotes elsewhere. An exported `type EvalMode` would cover it.
- The `messages.map((msg: { role: string; parts?: ... })` annotation asserts a shape nothing checks. A `[null]` element throws inside and the outer catch swallows it, landing on the fallback above. An explicit object guard reads better than relying on the catch.

PR #217, opened 2026-08-27
feat(dt-eval-cli): print resolved config path and carry eval name in bizevents (AI-459)

yassine-dt:
## What

Addresses two small pieces of feedback (AI-459):

- **Print the config file used by `dt-evals run`.** The command now resolves and prints the config path it picked up (`Using config: <path>`) before the run starts — suppressed in `--ci` mode.
- **Carry the eval name into bizevents.** Adds a `dt.eval.name` field to the evaluation-result bizevent, populated from the config's `name`, so results can be grouped/filtered by eval configuration.

## Verify

Run against real traces:
\`\`\`dql
fetch bizevents, from: now()-30m
| filter event.type == "gen_ai.evaluation.result"
| summarize count = count(), by: {\`dt.eval.name\`}
\`\`\`

Confirmed end-to-end: all written results carry the config's `name`, and the CLI prints the resolved config path.

### DIFF
--- a/dt-eval-cli/src/cli/commands/run.ts
+++ b/dt-eval-cli/src/cli/commands/run.ts
@@ -1,5 +1,5 @@
 import { Command } from 'commander';
-import { loadConfig, validateConfig } from '../../config/index.js';
+import { loadConfig, validateConfig, getProjectConfigPath } from '../../config/index.js';
 import { resolveEndpoints, metricId } from '../../config/schema.js';
 import { DynatraceClient } from '../../dt/client.js';
 import { runEvals, type RunProgressEvent, type RunResult } from '../../runner/index.js';
@@ -115,6 +115,7 @@ export function createRunCommand(): Command {
       configureLogger({ level: 'debug' });
     }
     const configPath = configArg ?? options.config;
+    const resolvedConfigPath = configPath ?? getProjectConfigPath();
     let config;
     try {
       config = loadConfig(configPath ? { projectFile: configPath } : undefined);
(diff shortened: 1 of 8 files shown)


dt-eval-cli/src/cli/commands/run.ts
```diff
@@ -115,6 +115,7 @@ export function createRunCommand(): Command {
       configureLogger({ level: 'debug' });
     }
     const configPath = configArg ?? options.config;
+    const resolvedConfigPath = configPath ?? getProjectConfigPath();
     let config;
     try {
       config = loadConfig(configPath ? { projectFile: configPath } : undefined);
```
thisthat, line 120: load config is a bit weird 🤔 https://github.com/dynatrace-oss/dt-evals/blob/e30cb1f210fd13c9e0814b2b28d3d7b3d87719ad/dt-eval-cli/src/config/index.ts#L168

if `configPath` is `null`, it will merge the config read from the global and local projects.

```
function getGlobalConfigPath(): string {
  return join(homedir(), '.dt-eval', 'config.yaml');
}

function getProjectConfigPath(): string {
  return join(process.cwd(), '.dt-eval.yaml');
}
``` 

Keeping this logic align in both places is prone to crack quickly. I suggest changing the config interface to store the file path from where it was read from.
This way, in `loadConfig` we first set it on the global, then we read the one from the project which ovverides the file path. 
Finally, we print in the log this new field of the interface

thisthat (approved):
:shipit:

PR #232, opened 2026-09-04
feat(dt-eval-cli): surface scope level and code evals in the configure wizard

yassine-dt:
## Summary

Surfaces two recently-added config capabilities in the interactive `dt-evals configure` wizard, which previously required hand-editing YAML:

1. **Scope level** — a `select` (agent-span / agent-session) so users can opt into session-level evaluation (grouped by `gen_ai.conversation.id`, falling back to trace id) without editing the file.
2. **Code evals (deterministic evals)** — a looping authoring menu that mirrors the LLM-judge evaluator list. Pick a method, fill in its mandatory fields, choose the span field to check, then return to the menu to add more or finish.

The metric checkbox is relabeled **"LLM-as-judge evaluators"** to distinguish it from the new code-evals section.

## Code evals flow

Menu of the six deterministic methods → per-method fields, each mapping 1:1 to an engine-honored param:

| Method | Fields collected |
|---|---|
| `must_contain` / `must_not_contain` | keywords (CSV), match mode any/all, case-sensitive |
| `exact_match` | expected output, case-sensitive, trim |
| `regex` / `must_not_match` | pattern, optional flags |
| `json_schema` | JSON object (validated on entry) |

Every entry also asks **which span field** to run against (output/input/context), written as `inputs.output` routing. Leaving the id blank discards the check and returns to the method list (**back**). Defaults match the engine's own defaults (`mode=any`, contains case-insensitive, exact-match case-sensitive).

Flag-only mode is unchanged — it preserves existing `scope.level` / metrics via the `existing` spread.

## Verification
- `npm run typecheck` — passes
- `npm run build` — passes (no `lint` script in this package)
- Every wizard option cross-checked against the deterministic engine (`contains.ts`, `exact-match.ts`, `regex.ts`, `json-schema.ts`) and the runner's `buildEvalInput`/`resolveCanonicalField` field routing — no dead options.

### DIFF
--- a/dt-eval-cli/package-lock.json
+++ b/dt-eval-cli/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@anthropic-ai/sdk": "^0.32.0",
         "@aws-sdk/client-bedrock-runtime": "^3.0.0",
-        "@dynatrace-oss/dt-eval-lib": "^0.0.15-alpha",
+        "@dynatrace-oss/dt-eval-lib": "^0.0.19-alpha",
         "@google/genai": "^1.0.0",
         "@inquirer/prompts": "^7.10.1",
         "@types/figlet": "^1.7.0",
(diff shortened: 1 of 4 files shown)


thisthat (approved):
LGTM 👍  
In a follow-up we could show how many of a specific code-eval entry were already configured. This way, when I am back at the `Code evals (deterministic evals) — 1 added; add another or finish` message , in the list I can see what I have already added. 
🤔  I dunno if it makes total sense though

PR #233, opened 2026-09-08
feat(dt-eval-cli): filter spans by arbitrary attributes

yassine-dt:
## Summary

- Adds `scope.filters` — a map of span attribute names to a value or list of values — compiled into the DQL fetch so filtering happens server-side in Grail.
- Multiple keys AND together; multiple values for one key OR together via `in()`.
- Validated in `validateConfig`, so a malformed filter fails at `dt-evals validate` instead of mid-run.

## Why

A run could only be scoped by service name, GenAI operation name, and an errors-only flag. That is not enough to isolate an orchestrator's routing spans when the orchestrator and its agents share a service, because every agent invocation emits the same `gen_ai.operation.name` (`invoke_agent`). This unblocks routing-accuracy evals (AI-496) without baking routing assumptions into the CLI.

```yaml
scope:
  filters:
    gen_ai.agent.name: router
    router.decision_type:
      - route
      - fallback
```

## Notes for review

- Attribute **values** are escaped with the existing `dqlStringLiteral()`. Attribute **keys** render as bare DQL identifiers, so they cannot be escaped — they are whitelisted against `/^[A-Za-z_][A-Za-z0-9_]*(\.[A-Za-z0-9_]+)*$/` and rejected otherwise. This is the injection boundary and the one comment in the diff.
- Clauses are emitted before `| fields`, so filtered attributes do not need adding to the projection list.
- Equality only — no `!=`, wildcards, or typed comparison. Values are always quoted strings, so a genuinely boolean/numeric attribute will not match. Deliberate; happy to follow up if wanted.
- Unlike `operationNames`, there is no parser-side safety net re-applying these filters after the fetch. Left out to keep the change minimal.

## Test plan

- [x] `npm test` — 300/300 passing
- [x] `npx tsc --noEmit` — clean
- [x] Rendered a query by hand to confirm clause ordering and `in()` output
- [ ] Validate against a real tenant with orchestrator spans

### DIFF
--- a/dt-eval-cli/src/dt/dql.ts
+++ b/dt-eval-cli/src/dt/dql.ts
@@ -64,17 +64,6 @@ function dqlStringLiteral(value: string): string {
   return `"${value.replace(/\\/g, '\\\\').replace(/"/g, '\\"')}"`;
 }
 
-// Attribute keys become raw DQL identifiers (unquoted), so unlike values they can't be
-// escaped. validateConfig rejects malformed keys first; this is a safety net at the
-// interpolation boundary so a future caller can't inject DQL by skipping validation.
-const ATTRIBUTE_KEY_PATTERN = /^[A-Za-z_][A-Za-z0-9_]*(\.[A-Za-z0-9_]+)*$/;
-
-function assertValidAttributeKey(key: string): void {
-  if (!ATTRIBUTE_KEY_PATTERN.test(key)) {
-    throw new Error(`Invalid span attribute key in scope.filters: "${key}"`);
-  }
-}
-
 /**
  * Resolve the effective operation-name keep-list, applying the default when unset and
  * trimming each entry so a stray-whitespace config value (e.g. `" chat "`) still matches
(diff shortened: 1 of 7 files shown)


dt-eval-cli/src/dt/dql.ts
```diff
@@ -62,6 +64,17 @@ function dqlStringLiteral(value: string): string {
   return `"${value.replace(/\\/g, '\\\\').replace(/"/g, '\\"')}"`;
 }
 
+// Attribute keys become raw DQL identifiers (unquoted), so unlike values they can't be
+// escaped. validateConfig rejects malformed keys first; this is a safety net at the
+// interpolation boundary so a future caller can't inject DQL by skipping validation.
+const ATTRIBUTE_KEY_PATTERN = /^[A-Za-z_][A-Za-z0-9_]*(\.[A-Za-z0-9_]+)*$/;
```
thisthat: I'd remove this or adapt the check :/ 
https://docs.dynatrace.com/docs/platform/grail/dynatrace-query-language/dql-reference#field-naming-rules
We could have something that is not following this in backticks 
As first version, I'd go with garbage-in/garbage-out. No checks, if they have something bad, no data will be returned and the server should give us the error about a wrong DQL

thisthat (commented):
We should add an example in the docs on how to use the new feature

dt-eval-cli/src/dt/dql.ts
yassine-dt: yes good point I will delete it

thisthat (approved):

PR #234, opened 2026-09-08
feat(dt-eval-cli): add evaluators --from-file and routing examples

yassine-dt:
## What

Ships two routing-accuracy eval examples — one per routing decision, one per conversation — plus a non-interactive way to install any evaluator definition from a JSON file.

## Why

Routing accuracy can't be a built-in evaluator: the routing policy *is* the agent roster, which differs per setup, so it has to ship as a template. But custom evaluators could only be created through the interactive wizard, meaning a shipped definition could only be installed by hand-editing `~/.config/dt-eval/custom-prompts.json`.

## Changes

- `evaluators add --from-file <path>` installs a JSON `PromptDefinition` non-interactively, validating it first since `createCustomPrompt` does not. A prompt may carry either placeholder, not both — `routing-accuracy` deliberately never sees the query, and `renderPrompt` already strips the unused one.
- `routing-accuracy` (`agent-span`) scores one handoff at a time against the sub-task it delegated. For single-shot routers.
- `session-routing-accuracy` (`agent-session`) scores a whole conversation's handoffs, so it can judge coverage and ordering. For multi-step planners.
- README explains which to pick, and documents the limitations found during validation.
- `.gitignore`: `*.dt-eval.yaml` was silently swallowing the example configs, so `examples/` is re-included, with a following `examples/**/.env*` rule so the negation can't resurrect a stray `.env`.
- Tests assert both shipped evaluators pass validation, so neither can rot.

## Validation notes

Both variants were run against a live 3-service orchestrator. `session-routing-accuracy` caught a genuine defect: a pricing agent invoked in parallel with the route planner on placeholder carrier data, instead of sequentially after it.


dt-eval-cli/examples/routing-accuracy/README.md
```diff
@@ -0,0 +1,91 @@
...
+> framework records handoffs on a dedicated `invoke_agent` span, that filter
+> won't match. First identify which span carries the routing decision, then
+> point `scope.filters` at it. To see your own distribution:
+>
+> ```dql
+> fetch spans
+> | filter isNotNull(gen_ai.provider.name)
+> | summarize count(), by:{gen_ai.operation.name, gen_ai.response.finish_reasons}
+> ```
+
```
thisthat, line 111: I would add here a `## Results` section with an example of the results. This could also be a gif or a c/p of the results of the command run

thisthat (approved):

PR #239, opened 2026-09-09
fix(dt-eval-cli): keep execute_tool spans in the genai-span filter

yassine-dt:
## What

The GenAI span-fetch query now also keeps spans carrying `gen_ai.tool.name`, so `execute_tool` spans become fetchable.

## Why

The fetch filter gated on `gen_ai.system` or `gen_ai.provider.name`. `execute_tool` spans carry neither — only `gen_ai.tool.name` — so a config scoped to `operationNames: [execute_tool]` matched zero spans, blocking any deterministic eval of tool-call arguments or results.

## Changes

- Broaden the guard in `buildGenAiSpanQuery` to also accept `isNotNull(gen_ai.tool.name)`.
- Add coverage in `dql.test.ts` (guard accepts tool.name; a config scoped to `execute_tool` reaches those spans).
- No effect on existing configs: the default operation keep-list excludes `execute_tool`, so tool spans are only fetched when a config explicitly opts in.


yassine-dt:
In tool call span neitehr provider.name or genai.system are required --> tool call spans are skipped in many tennats for example: 
**BB**
Screenshot 2026-09-09 at 16 45 08
**XBW  --> we have provider.name** 
image

**GUU demo tenant**
image

thisthat (approved):

PR #240, opened 2026-09-09
feat(dt-eval-cli): add tool-call-accuracy example templates

yassine-dt:
## What

A new `examples/tool-call-accuracy/` folder documenting how to evaluate a leaf tool call two ways: deterministic checks on its arguments/result, and an LLM judge on the tool-choice decision.

## Why

A tool call fails in three separable ways — malformed arguments, an unproductive (empty) result, or a well-formed call that is simply wrong for the request. The first two are rules; only the third needs a judge. No existing example covers this leaf-tool layer (`routing-accuracy` judges which *agent* a request goes to, not a call's *arguments*).

## Changes

- `tool-call-checks.dt-eval.yaml` — deterministic, no LLM: `json_schema` over `gen_ai.tool.call.arguments` (well-formed) + `must_not_match` over the result text (returned data), via `spanFields` plus per-metric `inputs` routing.
- `tool-call-accuracy.dt-eval.yaml` + `.evaluator.json` — custom LLM judge scoring tool choice and argument fidelity against the delegated task.
- `README.md` — when to use each, the `execute_tool` span mapping, and real run output.
- Screenshots pending — `images/tool-call-checks-results.png` and `images/tool-call-accuracy-results.png` to be added before merge.

Merge after #239 — the deterministic config reads `execute_tool` spans, which the current span-fetch guard drops; #239 adds `gen_ai.tool.name` to it. Without #239, `tool-call-checks` returns 0 spans.


dt-eval-cli/examples/tool-call-accuracy/README.md
```diff
@@ -0,0 +1,184 @@
...
+judge for the one that actually needs reasoning.
+
+## Which check to use
+
+| | `tool-call-checks` | `tool-call-accuracy` |
+|---|---|---|
+| Kind | deterministic, no LLM | LLM judge |
+| Span read | `execute_tool` — the tool call itself | the agent's LLM span — the decision to call it |
+| Answers | are the arguments well-formed? did the tool return data? | was the right tool chosen, with values faithful to the request? |
+| Cost | none | one judge call per step |
```
yassine-dt: Change cost it is not accuarte

thisthat (approved):
:shipit:

PR #245, opened 2026-09-17
feat(dt-eval-lib): add per-metric score direction attribute

yassine-dt:
## What

Adds an optional `ScoreDirection` (`positive` | `negative` | `neutral`) to each metric definition, and emits it on evaluation bizevents, so consumers can distinguish risk-named metrics (toxicity, bias, …) from quality-named ones.

## Why

Risk metrics display as contradictory (e.g. a passing Toxicity result shown as a high green score reads as "100% toxic"). A per-metric direction lets downstream consumers reframe display without changing scoring. Scores stay canonical (high = pass) so this is purely additive metadata.

## Changes

- Add `ScoreDirection` type and optional `direction` field on `PromptDefinition`; export it from the public API.
- Tag the 7 risk metrics `negative` and the 8 quality metrics `positive` in the catalog; absent defaults to higher-is-better.
- Emit `gen_ai.evaluation.direction` on evaluation bizevents, threaded from the prompt definition (defaults to `positive` when absent).
- Add guard tests for catalog directions and bizevent emission.
- No behavioral change: `computeScore`, thresholds, and pass/fail are untouched.

Note: UI consumption of `direction` is a separate follow-up.


dt-eval-cli/src/runner/index.ts
```diff
@@ -130,6 +130,7 @@ interface EvalTaskResult {
   span: GenAiSpan;
   metricId: string;
   metricName: string;
+  direction?: ScoreDirection;
```
stefan-falk-dt, line 133: Without having the context here, so my apologies if that suggestion does not make sense:

Sounds like this should communicate the sentiment? 

More precisely, the polarity? Would it make sense to have something like:

- `sentimentPolarity: 'positive' | 'negative' | 'neutral';`
- `sentiment: { polarity: 'positive' | 'negative' | 'neutral' };`

Note: One could also use numbers `-1, 0, 1` if we ever wanted to provide a real-valued score later :)

yassine-dt: Thanks stefan for the review 
You can get the context here https://dynatrace.enterprise.slack.com/archives/C0ACSJD3M6Z/p1789633923435739 
It is not about the sentiment but what is bad. For now we have a convention lower is bad 0 --> pass is 1 
But this is contradictory with hallucination for example we want the opposite.
So I am adding this field so the UI can decide if negative flip the scale 1- actual value if not keep it.
Let me know if this make sense.

yassine-dt: https://dynatrace.enterprise.slack.com/archives/C08PCHUKQEA/p1789603185457259 
More context from the help channel

stefan-falk-dt: Ah, I see. makes sense. :) 

If it's a public API and this field is supposed to explain how the score has to be interpreted, then it could be an option to go with something like `higher_is_better | lower_is_better | neutral`.

But that's a nit-pick at this point 😄

